PR #3, opened 2024-11-22
Feature/api route

Iarynovskyi:
added flask-api + cache


api/routes/__init__.py
```diff
@@ -1,17 +1,22 @@
...
-from api.routes import api_news, api_games, api_sports
+from api.routes import api_news, api_games, api_sports, api_teams
+from api.routes.cache import cache
 
 # Ініціалізація додатку
 def create_app():
     app = Flask(__name__)
     CORS(app)
+    app.config['CACHE_TYPE'] = 'SimpleCache'
+    app.config['CACHE_DEFAULT_TIMEOUT'] = 300
```
Avariq, line 11: What are those? seconds? please specify.

If those are 5 minutes, I'd suggest writing 60 * 5 and writing a comment "5 min" or smth

api/routes/api_games.py
```diff
@@ -1,14 +1,31 @@
...
+
+
+@games_app.route('/<sport_type>', methods=['GET'])
+@cache.cached(60*1.3, key_prefix=make_cache_key)
+def get_sport_stream_info_today_endpoint(sport_type):
+    try:
+       games = get_stream_info_for_sport(session, sport_type)
+       return games, 200
+    except Exception as e:
+       return jsonify({"error": str(e)}), 500
```
Avariq, line 31: if it's used more than once consider separating it into a different method e.g. "get_error_response" or whatever

api/routes/api_news.py
```diff
@@ -1,18 +1,31 @@
...
-from service.api_logic.news_logic import get_news_by_count
+from flask import Blueprint, jsonify, request
+from service.api_logic.news_logic import get_news_by_count, get_latest_sport_news
 from database.session import SessionLocal
+from api.routes.cache import cache
 
 session = SessionLocal()
 news_app = Blueprint('news', __name__)
 
+def make_cache_key():
```
Avariq, line 9: can we only have one of those decorators in our code? If it serves the exact same purpose as the one in api_games file we should reuse it by moving it to some other place first and then importing wherever needed.

api/routes/api_teams.py
```diff
@@ -0,0 +1,42 @@
+from flask import Blueprint, jsonify, request
+from database.session import SessionLocal
+
+session = SessionLocal()
+teams_app = Blueprint('teams', __name__)
+
+def make_cache_key():
```
Avariq, line 7: same here

requirements.txt
```diff
@@ -0,0 +1,10 @@
+Flask-Caching~=2.3.0
```
Avariq, line 1: about damn time

service/api_logic/news_logic.py
```diff
@@ -1,13 +1,12 @@
 import json
-from database.models import News
+from database.models import News, Sport
 from database.azure_blob_storage.save_get_blob import blob_get_news
 
 
 def get_news_by_count(count: int, session) -> str:
     news_records = session.query(News).order_by(News.save_at.desc()).limit(count).all()
     if not news_records:
         print(f"\033[31mNo news was found in the database.\033[0m")
```
Avariq, line 9: Consider creating separate method "print_text_in_red" (don't copy the method name. It's definitely not what you might want to have) which would handle the coloring of the text. 

Google DRY, YAGNI, KISS and, of course, SOLID.

service/api_logic/news_logic.py
```diff
@@ -18,4 +17,29 @@ def get_news_by_count(count: int, session) -> str:
...
         except Exception as e:
             print(f"\033[31mError while receiving blob '{news_record.blob_id}': {e}\033[0m")
+
     return json.dumps(all_results, ensure_ascii=False)
+
+
+def get_latest_sport_news(count: int, sport_name, session):
+    sport = session.query(Sport).filter(Sport.sport_name == sport_name).first()
+    if not sport:
+        return {"error": f"Sport '{sport_name}' not found"}
```
Avariq, line 27: Create logic for proper exception/error handling. Less hardcode.

You might wanna look into "Custom Exceptions" on Google, or at least create a method to form the body of the error

Avariq (changes requested):
Please fix the issues before merging

requirements.txt
Iarynovskyi: what's wrong here ?

api/routes/api_teams.py
Avariq, line 7: Probably missed this one :)

Avariq (approved):
MINOR: api_teams is till declaring "make_cache_key" function.

PR #5, opened 2024-12-05
Feature/main page frontend

AndrewKozovyi:
UJP-21
UJP-85


frontend/src/pages/mainPage.js
```diff
@@ -147,8 +147,8 @@ function MainPage() {
                     {sports.map((item, index)=>(
                         <SportBtn
                             key={index}
-                            sport={item.data?.sport_name}
-                            img={item.data?.sport_img}
```
Iarynovskyi: Some durak

frontend/src/pages/mainPage.js
```diff
@@ -1,32 +1,57 @@
...
+    const [games, setGames] = useState([]);
+
+    useEffect(() => {
+        axios.get(`${apiEndpoints.url}${apiEndpoints.news.getRecent}`)
+            .then(res => {
+                const returnedNews = res.data;
+                setNews(returnedNews);
+            })
+            .catch(error => {
+                console.error('There was an error getting news:', error);
```
Avariq, line 24: Think about implementing some sort of a pop-up (any kind will do) to display those errors to the end user.

Regular customer does not know what console is let alone how to open it

frontend/src/pages/mainPage.js
```diff
@@ -39,28 +64,22 @@ function MainPage({news}) {
...
-                                <h4 className="online">етер</h4>
-
-                                <h1 className="score">0:0</h1>
-
-                                <h4 className="matchLeague">ліга</h4>
-
-                            </div>
-
-                            <i className="fa fa-user-o" aria-hidden="true"></i>
+                        {games.slice(0, 5).map((item, index) => (
```
Avariq, line 69: Hardcoded values are now welcome unless crucial and required  (I saw a couple places with same values used in the same way).
Consider moving it to some global variable or something like that

frontend/src/pages/sportPage.js
```diff
@@ -0,0 +1,110 @@
...
+                    />
+                ))}
+
+            </section>
+
+            <ReactPaginate
+                breakLabel="..."
+                nextLabel="→"
+                onPageChange={handlePageClick}
+                pageRangeDisplayed={3}
```
Avariq, line 98: I'd also consider having that separated into some variable.

Maybe in future you will want a bigger number here (for example for people with big screens)

frontend/src/pages/mainPage.js
AndrewKozovyi: That was a test slice by @Romlen2, he's trying to get correct path to the required data

frontend/src/pages/mainPage.js
AndrewKozovyi: Done

frontend/src/pages/sportPage.js
AndrewKozovyi: I guess, it's done for now

Avariq (approved):
Approve for now, but pop-up logic should be custom-implemented in future

(Default alerts are sooo 2007...)

PR #6, opened 2024-12-05
Feature/api route

Iarynovskyi:
UJP-86, UJP-126


AndrewKozovyi:
♥

api/routes/api_teams.py
```diff
@@ -1,35 +1,67 @@
...
+
 @teams_app.route("/", methods=['GET'])
+@cache.cached(timeout=60*60)
 def get_teams_endpoint():
     try:
         all_teams = get_teams(session)
         return all_teams
     except Exception as e:
         print(e)
+    except OperationalError as e:
```
Avariq, line 40: this one is never getting executed, because Exception is a base class and catching it means catching every single exception there is

change the except chain order

Avariq, line 40: also in the code above that I see decorators handling the exceptions while here it is manual.

you must figure out one best approach and follow it. 
keep the code as clean as possible

service/api_logic/games_logic.py
```diff
@@ -21,7 +21,9 @@
 def filter_matches_by_date(matches, today, date_key=DATE_KEY):
```
Avariq, line 21: I see the default value being set to 'DATE_KEY', but I can hardly see any potential usage for that

service/api_logic/teams_logic.py
```diff
@@ -8,27 +14,45 @@
...
+    return None
 
-    return None
\ No newline at end of file
+
+def get_teams_sport(session, sport_type):
+    try:
+        sport = get_sport_by_name(session, sport_type)
+    except SportNotFoundError as e:
+        print_error_message({"error": e.message})
```
Avariq, line 47: {"error": e.message} is duplicated

store in variable and reuse

service/implementation/auto_request_api/__init__.py
```diff
@@ -1,29 +1,41 @@
 import time
 from apscheduler.schedulers.background import BackgroundScheduler
 from concurrent.futures import ThreadPoolExecutor
-from logic_auto_request import apis, auto_request_system
+from service.implementation.auto_request_api.logic_auto_request import apis, auto_request_system, keep_db_alive
```
Avariq, line 4: Placing executable code (and especially something like here) is generally a bad idea:

"The __init__.py file is intended for package initialization, not for running or scheduling tasks. It should primarily set up the namespace for a package or perform lightweight configuration."

__init__.py is sort of a header for your package and it gets executed every time this package is imported somewhere. It should describe/prepare the package for correct workflow instead of executing real logic.

I strongly suggest changing this logic

api/routes/api_games.py
```diff
@@ -16,17 +16,12 @@ def handle_exception(e):
 @games_app.route('/', methods=['GET'])
 @cache.cached(60*1.3)
 def get_stream_info_today_endpoint():
-   try:
-      games = get_stream_info_today(session)
-      return games
-   except Exception as e:
-       print(e)
+    games = get_stream_info_today(session)
+    return games
```
Avariq, line 20: why did we remove try-except here?

api/routes/api_games.py
```diff
@@ -16,17 +16,12 @@ def handle_exception(e):
...
 @games_app.route('/<sport_type>', methods=['GET'])
 @cache.cached(60*1.3, key_prefix=make_cache_key)
 def get_sport_stream_info_today_endpoint(sport_type):
-    try:
-       games = get_stream_info_for_sport(session, sport_type)
-       return games
-    except Exception as e:
-        print(e)
\ No newline at end of file
+    games = get_stream_info_for_sport(session, sport_type)
```
Avariq, line 26: same here. try-except?

api/routes/api_sports.py
```diff
@@ -11,10 +12,15 @@ def handle_exception(e):
...
+@sports_app.route('/league/<sport_type>', methods=['GET'])
+@cache.cached(timeout=60*60)
 def get_all_sports_endpoint():
-    try:
-        sports = get_all_sports()
-        return sports
-    except Exception as e:
-        print(e)
\ No newline at end of file
+    sports = 1#get_all_leagues_by_sport(session)
```
Avariq, line 25: should this be commented?

api/routes/api_teams.py
```diff
@@ -12,41 +15,50 @@ def handle_exception(e):
...
-        return response
-    except Exception as e:
-        print(e)
+    data = request.get_json()
+    response = {
+        "sport": data.get("sport", "Unknown"),
+        "team_id": data.get("team_id", "Unknown"),
+        "league_id": 3
+    }
+    response = rugby_teams_statistics(response)
```
Avariq, line 50: that is now very SOLID, because your first "response" serves as an input to your following method execution.

It'd be better to just do response = rugby_teams_statistics(input_parameters)

I also remind you to use predefined models and DTOs to transfer such data (you can have one for input and different one for output). Everywhere I look I see just plain dictionaries etc etc: it's not convenient and it hinders debug process for whoever decides to do debugging in the future.. Think of a better way to do this

api/routes/api_teams.py
```diff
@@ -12,41 +15,50 @@ def handle_exception(e):
...
-        #response = players_info()
-        return response
-    except Exception as e:
-       print(e)
+    data = request.get_json()
+    response = {
+        "sport": data.get("sport", "Unknown"),
+        "team_id": data.get("team_id", "Unknown"),
+    }
+    response = basketball_players(response)
```
Avariq, line 61: Same here: output is used as input

service/api_logic/news_logic.py
```diff
@@ -20,24 +19,29 @@ def fetch_news(session, order_by: ClauseElement = None, limit: int = None, filte
 
 
 def get_news_by_count(count: int, session):
-    news = fetch_news(session, order_by=News.save_at.desc(), limit=count)
-    return json_news(news)
+    try:
+        news = fetch_news(session, order_by=News.save_at.desc(), limit=count)
+        return json_news(news)
+    except Exception:
```
Avariq, line 25: except-raise again

exception won't be handled this way.

By the way: I wonder what would happen if I pass negative number here :)

Do validation of the incoming data. That is usually done in the API layer

service/api_logic/news_logic.py
```diff
@@ -20,24 +19,29 @@ def fetch_news(session, order_by: ClauseElement = None, limit: int = None, filte
...
     return json_news(news)
 
 
 def get_popular_news(count: int, session):
-    news = fetch_news(session, order_by=News.interest_rate.desc(), limit=count)
-    return json_news(news)
+    try:
+        news = fetch_news(session, order_by=News.interest_rate.desc(), limit=count)
+        return json_news(news)
+    except Exception:
```
Avariq, line 43: except-raise again

Also, catching "Exception" (base class) here is meaningless, since we don't know how to handle it. It basically means "Something wrong happened" so it's better moved to API's endpoints.

Here, on the other hand, we should only catch exceptions we expect (if any). For instance, let's say you have an algorithm that traverses a list of items by their indexes. You expect that algorithm might not work in certain cases and Some IndexOutOfRange exception might occur. In this case we will catch that paticular exception (try-except IndexOutOfRange as e) and do some custom logic if required or use another algorithm - whatever. Thing is, that we know that this exception might happen and we know how to deal with it - it does not break our logic. Such exceptions are worth catching here, otherwise - catch Exception on the endpoint 

FYI:
by re-raising this Exception you not only allow your app to crush, but also have a good chance of removing original exception's message

service/api_logic/sports_logic.py
```diff
@@ -1,17 +1,35 @@
 from database.models import Sport
-from database.session import SessionLocal
 from exept.colors_text import print_error_message
+#from database.models.league import League
```
Avariq, line 3: commented import

service/api_logic/sports_logic.py
```diff
@@ -1,17 +1,35 @@
...
+        if not sports:
+            print_error_message("No sports found in the database")
+            return [], 204
+        return [
+            {
+                "sport_id": sport.sport_id,
+                "sport_name": sport.sport_name,
+                "sport_img": sport.sport_img
+            } for sport in sports], 200
+    except Exception as e:
```
Avariq, line 17: except-raise again

service/api_logic/sports_logic.py
```diff
@@ -1,17 +1,35 @@
...
-        print_error_message("No sports found in the database")
-        return [], 204
-    return [
-        {
-            "sport_id": sport.sport_id,
-            "sport_name": sport.sport_name,
-            "sport_img": sport.sport_img
-        } for sport in sports], 200
\ No newline at end of file
+'''
```
Avariq, line 20: commented code


BTW, I'm pretty sure that is not the proper way of commenting in python. You should use '#' on each new line. You can do that by using CTRL+/ 9by default in most code editors) after selecting the code you're willing to comment

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
 RESPONSE_KEY = "response"
 
 def get_teams(session):
-    blob_indexes = get_all_blob_indexes_from_db(session, TEAMS_JSON)
-    result = get_blob_data_for_all_sports(session, blob_indexes)
-    data = json.loads(result)
-    filtered_data = []
+    try:
+        blob_indexes = get_all_blob_indexes_from_db(session, TEAMS_JSON)
```
Avariq, line 16: variable naming

if we expect blob indexes for teams - then specify it: teams_blob_indexes or blob_indexes_teams

Do it for every variable you write (I'm not asking you to rewrite every variable name here,. Just telling you that for future)

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
+        result = get_blob_data_for_all_sports(session, blob_indexes)
```
Avariq, line 17: sport_result

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
+        data = json.loads(result)
```
Avariq, line 18: sport_result_json

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
-            filtered_data.append(processed_data)
-    return filtered_data, 200
+        for teams_data in data:
+            processed_data = process_blob_data(teams_data)
+            if processed_data:
+                filtered_data.append(processed_data)
+        return filtered_data, 200
+    except OperationalError:
+        raise DatabaseConnectionError()
+    except Exception:
```
Avariq, line 28: except-raise again

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
     team = sport_data.get(DATA_KEY, {}).get(RESPONSE_KEY, [])
     sport = sport_data.get(SPORT_KEY)
     if team:
         return {
-            "sport": sport,
-            #"blob_name": blob_name,
+            SPORT_KEY: sport,
             TEAM_KEY: team,
         }
+    return None
```
Avariq, line 40: I hope you guys handle this 'None' somewhere

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
+        filtered_data = []
+        for teams_data in data:
+            if sport_type == teams_data.get(SPORT_KEY):
+                processed_data = process_blob_data(teams_data)
+                if processed_data:
+                    filtered_data.append(processed_data)
+        return filtered_data, 200
+    except OperationalError:
+        raise DatabaseConnectionError()
+    except Exception:
```
Avariq, line 61: except-raise again

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
-    return None
\ No newline at end of file
+
+def get_teams_sport(session, sport_type):
+    try:
+        sport = get_sport_by_name(session, sport_type)
+    except SportNotFoundError as e:
+        return get_error_response({"error": e.message}, 404)
+    try:
+        blob_indexes = get_all_blob_indexes_from_db(session, TEAMS_JSON)
```
Avariq, line 49: varables

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
+        result = get_blob_data_for_all_sports(session, blob_indexes)
+        data = json.loads(result)
+        filtered_data = []
+        for teams_data in data:
```
Avariq, line 53: while it's not a mistake to write code like this, it is still not "Python-like".

use your programming language techniques. For example the code above could be rewritten to:

filtered_data = [
    processed_data
    for teams_data in data
    if sport_type == teams_data.get(SPORT_KEY) and (processed_data := process_blob_data(teams_data))
]

should work, but, of course, I didn't test it :)

service/api_logic/teams_logic.py
```diff
@@ -8,27 +12,94 @@
...
+                if processed_data:
+                    filtered_data.append(processed_data)
+        return filtered_data, 200
+    except OperationalError:
+        raise DatabaseConnectionError()
+    except Exception:
+        raise Exception
+
+
+"-------------------TEST--------------------"
```
Avariq, line 65: It's not a good practice to commit test/debug code to the branch.

If your task is finished - PR should not have those test/debug/commented lines. And if your task is not finished - you should not have opened the PR in the first place. 

That's the "Code of conduct" in programming, that we use

service/implementation/auto_request_api/logic_auto_request.py
```diff
@@ -38,20 +42,23 @@
         "host":"v3.football.api-sports.io",
         "frequency": 1335
     },
+    #--------------------------------
```
Avariq, line 45: not sure if thse needed here, but it's up to you

service/implementation/auto_request_api/logic_auto_request.py
```diff
@@ -425,4 +440,13 @@ def auto_request_system(api: Dict[str, str]) -> None:
...
+        print(f"Загальна помилка при збереженні даних для {api['name']}: {e}")
+
+
+def keep_db_alive():
+    session = scoped_session(SessionLocal)
+    try:
+        session.execute(text("SELECT 1"))
+        session.commit()
+    except Exception as e:
+        print(f"Keep-alive failed: {e}")
```
Avariq, line 452: some text/exceptions is/are in Ukrainian and others in English - that should be fixed. Use one style and language throughout all your project files

service/implementation/auto_request_api/logic_request_by_react.py
```diff
@@ -172,6 +172,7 @@ def afl_teams_statistics(api_data: Dict[str, str]) -> Dict[str, str]:
 
 
 def afl_players(api_data: Dict[str, str]) -> Dict[str, str]:
+    print(api_data)
```
Avariq, line 175: that also belongs to "debug" code, which should not be present on the github.

I'll tell you about logging on Friday

service/implementation/auto_request_api/logic_request_by_react.py
```diff
@@ -505,7 +506,7 @@ def volleyball_teams_statistics(api_data: Dict[str, str]) -> Dict[str, str]:
     except Exception as e:
         return {"error": str(e)}
 
-
+'''
```
Avariq, line 509: commented code again (and incorrectly commented as well - use '#')

api/routes/api_games.py
Iarynovskyi: because I moved all the logic under the hood into functions
I thought it would be better that way

api/routes/api_sports.py
Iarynovskyi: so because the func doesn't work yet, it should be completed later in another task

api/routes/api_teams.py
Iarynovskyi: I have made UniversalResponseDTO for now
And later when Vlad finishes refactoring all those terrible functions,
I will finish the output for them

service/api_logic/teams_logic.py
Iarynovskyi: In theory, everything is good, that if there is nothing, it will be none and the array will return empty

service/api_logic/teams_logic.py
Iarynovskyi: My bad
Everything that I marked as a test was for other tasks
And since the develop doesn't have what I need yet
I started testing here, so now I'll just transfer it to the notes for now

api/routes/api_games.py
Iarynovskyi: The decorator will work either if there is an exception directly in the flask
Or if the corresponding errors are raised in the function
So i decided to do it under the hood
And thats why i do except-raise again and again

service/api_logic/teams_logic.py
Iarynovskyi: for error decorator in flask

service/api_logic/sports_logic.py
Iarynovskyi: Preparation for the future

service/api_logic/sports_logic.py
```diff
@@ -1,10 +1,9 @@
...
+from exept.handle_exeptions import handle_exceptions
 
-session = SessionLocal()
-
-def get_all_sports():
+@handle_exceptions
+def get_all_sports(session):
     sports = session.query(Sport).all()
     if not sports:
         print_error_message("No sports found in the database")
```
Avariq, line 9: It'd be better to raise a NotFound Exception here (you already have one. Some tuning and it will work here too)

Also, service layer (Business logic) does not care about API. It must handle its own tasks and that's it. So do not return status code here - it should be handled on API level (plus: if you raise a custom exception here there won't be any need for that return statement whatsoever)

And I'm not sure about 204 status code. It probably must've been 404.

Summary:
Throw a custom exception

service/api_logic/sports_logic.py
```diff
@@ -14,4 +13,32 @@ def get_all_sports():
...
+        session.query(League)
+        .join(Sport, Sport.sport_id == League.sport_id)
+        .filter(Sport.sport_name == sport_name)
+        .limit(per_page)
+        .offset(offset)
+        .all()
+    )
+    if not leagues:
+        print_error_message("No leagues found for the sport")
+        return [], 204
```
Avariq, line 34: Same here

Do not return status code.
Since we already went with custom exceptions approach - let's use it. 
Throw a custom exception here as well

service/api_logic/sports_logic.py
```diff
@@ -14,4 +13,32 @@ def get_all_sports():
...
             "sport_img": sport.sport_img
-        } for sport in sports], 200
\ No newline at end of file
+        } for sport in sports], 200
+
+
+@handle_exceptions
+def get_all_leagues_by_sport(session, sport_name, page, per_page):
+    if sport_name is None or "Unknown":
+        return {"error": "Not param sport_name"}
```
Avariq, line 22: Hardcoded dictionary.

We should either create a model/class to keep it or, again, use a custom exception.

Whatever the choice - do NOT hardcode strings

service/api_logic/sports_logic.py
```diff
@@ -14,4 +13,32 @@ def get_all_sports():
...
+        .filter(Sport.sport_name == sport_name)
+        .limit(per_page)
+        .offset(offset)
+        .all()
+    )
+    if not leagues:
+        print_error_message("No leagues found for the sport")
+        return [], 204
+    return [
+        {
```
Avariq, line 36: why don't we create a model/class to return this data?

then we would be able to do something like: myModel._id; myModel.name; myModel.logo... etc

This will make it easier to track/change later

service/api_logic/teams_logic.py
```diff
@@ -1,34 +1,60 @@
...
-    filtered_data = []
+        teams_blob_indexes = get_all_blob_indexes_from_db(session, TEAMS_JSON)
+        sport_result = get_blob_data_for_all_sports(session, teams_blob_indexes)
+        sport_result_json = json.loads(sport_result)
+        filtered_data = [
+            processed_data
+            for teams_data in sport_result_json
+            if (processed_data := process_blob_data(teams_data))
+        ]
+        return filtered_data, 200
```
Avariq, line 27: Status code again. Remove.

By the way, if you see multiple places with same return statement (in your case it would be: "return something, status_code") it's a good sign you need a model/class/DTO created.

Follow DRY principle (and SOLID, of course)

service/api_logic/teams_logic.py
```diff
@@ -1,34 +1,60 @@
...
+        teams_by_sport_blob_indexes = get_all_blob_indexes_from_db(session, TEAMS_JSON)
+        result = get_blob_data_for_all_sports(session, teams_by_sport_blob_indexes)
+        data = json.loads(result)
+
+        filtered_data = [
+            processed_data
+            for teams_data in data
+            if dto.sport == teams_data.get(SPORT_KEY) and (processed_data := process_blob_data(teams_data))
+        ]
+        return filtered_data, 200
```
Avariq, line 56: Status code again

api/routes/dto.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel
+from typing import Optional, Union
+
+class UniversalResponseDTO(BaseModel):
```
Avariq, line 4: MINOR: 
It's alright, but "Universal" in the name always triggers me. Maybe we can rename it somehow

service/api_logic/sports_logic.py
Iarynovskyi: I do this here because sometimes db can just be empty

api/routes/api_sports.py
```diff
@@ -1,20 +1,40 @@
...
+        return sports
+    except Exception as e:
+        response = {"error in service": str(e)}
+        return get_error_response(response, 500)
+
+@sports_app.route('/league', methods=['POST'])
+@cache.cached(timeout=60*60)
+def get_all_leagues_endpoint():
+    try:
+        data = request.get_json()
```
Avariq, line 32: I see a lot of places like this, where you parse the incoming data manually.
Use DTOs instead. They will handle all the validation and mapping for you (read what mapping to DTO is in API (ChatGPT?))

I am most familiar with Marshmallow library, which provides that functionality, so you can use it

Avariq: And, of course, this should be done for each endpoint we have   :)

**Quote by Aeschylus: “Only through suffering do we learn”**

Iarynovskyi: I have already written DTO models
But they are in the test branch
Otherwise, this PR will be constantly updated and not close never)

Avariq: Nice. (Though I'd like to merge only finished code into develop branch, but let it be)

Then fix other comments (if not already) and I'll approve

service/api_logic/sports_logic.py
Avariq: @Iarynovskyi 

I understand, do the validation for that. But my concern is your "return"

It'd be better to just:

`if not leagues:
   raise SomeCustomExceptionWithCorrectStatusCode(your_message)`

api/routes/scripts.py
```diff
@@ -1,10 +1,28 @@
 from flask import jsonify, request
-
+from flask import request
+import hashlib
+import json
 
 def get_error_response(e, status_code):
     return jsonify(e), status_code
 
+def get_good_response(result):
```
Avariq, line 9: Won't flask itself return a proper success status code if we just do "return results"?

exept/exeptions.py
```diff
@@ -1,23 +1,46 @@
-class SportNotFoundError(Exception):
+class SoftServeException(Exception):
```
Avariq, line 1: This should be renamed to something that is project-specific

Any general name will do better that "SoftServe"

Avariq (approved):
I still see places with plain dicts instead of DTOs and pydantic lib could be used to validate data even on higher levels (like allowing only positive numbers through) but otherwise looks better now

PR #7, opened 2024-12-10
Feature/data base architecture

Iarynovskyi:
UJP-118, sqlalchemy tables for new db and test funk for return games info from db with params


service/api_logic/games_logic.py
```diff
@@ -96,3 +99,88 @@ def get_stream_info_for_sport(session, sport_name):
...
+            TeamIndex.name.label("away_team_name"),
+            TeamIndex.logo.label("home_team_logo"),
+        )
+        .join(League, Games.league_id == League.league_id)
+        .join(Country, Games.country_id == Country.country_id)
+        .join(TeamIndex, Games.team_home_id == TeamIndex.team_index_id)
+        .join(TeamIndex, Games.team_away_id == TeamIndex.team_index_id)
+    )
+    filters = []
+    if sport_id is not None:
```
Avariq, line 135: should be done somewhere else

PR #8, opened 2024-12-20
Feature/UJP-119 - save data from api

torii044:
UJP-119


database/postgres/postgres_country.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if isinstance(data, list):
+        for item in data:
+            result = extract_country(item, key)
+            if result is not None:
+                return result
+
+    return None"""
+
+def save_country(country_data: Dict, session) -> Country:
```
Avariq, line 24: country_data should be some model/class - not a dictionary.

database/postgres/postgres_country.py
```diff
@@ -0,0 +1,54 @@
...
+
+    country_api_id = country_data.get('id')
+    country_name = country_data.get('name')
+    country_code = country_data.get('code')
+    country_flag = country_data.get('flag')
+
+    if not country_name:
+        return None
+
+    country_entry = session.query(Country).filter_by(name=country_name).first()
```
Avariq, line 39: Not sure but isn't there something like "first_or_default" in sqlalchemy?

I'm cautious about this line yielding unwanted exception. Not sure though, will dicuss tomorrow.

database/postgres/postgres_game.py
```diff
@@ -0,0 +1,102 @@
+from typing import Dict
+from datetime import datetime
+from database.models import Games
+from database.postgres.postgres_country import save_country
+from database.postgres.postgres_team import save_team
+from database.postgres.postgres_league import save_league
+
+def save_games(json_data: Dict, sport_id, session) -> None:
```
Avariq, line 8: json_data must be replaced with a model/class representing "games" list or something like that

database/postgres/postgres_game.py
```diff
@@ -0,0 +1,102 @@
...
+    games = json_data.get("response", [])
+    if not games:
+        print('insufficient data')
+        return
+
+    for game in games:
+        if "fixture" in game or "game" in game:
```
Avariq, line 15: DAL (DataAccessLayer) should not be handling such logic (parsing something).

The general, most wanted flow is:
`I get a model -> I verify a model -> I CREATE/UPDATE/DELETE the model entry from the database.`

database/postgres/postgres_game.py
```diff
@@ -0,0 +1,102 @@
...
+            separated_data = game.get("fixture") or game.get("game")
+            api_id = separated_data.get("id")
+            date = separated_data.get("date")
+            status = separated_data.get('status')
+        else:
+            api_id = game.get('id')
+            date = game.get('date')
+            status = game.get('status')
+
+        country_data = game.get("location").get("country") if "location" in game else game.get("country")
```
Avariq, line 25: We will replace it with a model/class anyway, but:

whenever you do something like here you MUST first verify whether such keys are present in your dictionary. What would happen if dictionary did not contain "country" within it? 

Our code has to be as "bulletproof" as possible

database/postgres/postgres_game.py
```diff
@@ -0,0 +1,102 @@
...
+        else:
+            time = game.get('time')
+
+        try:
+            if isinstance(date, str):
+                parsed_date = datetime.fromisoformat(date)
+            elif isinstance(date, datetime):
+                parsed_date = date
+            else:
+                raise ValueError("Невідомий формат дати")
```
Avariq, line 40: I believe we've decided to make all text/exceptions/messages written in English

database/postgres/postgres_league.py
```diff
@@ -0,0 +1,44 @@
+from typing import Dict
+from database.models import League
+from database.postgres.postgres_country import save_country
+
+def save_league(league_data, session, sport_id) -> League :
```
Avariq, line 5: model/class here as well

and also don't forget about datatypes (e.g. league_data: dict)

database/postgres/postgres_team.py
```diff
@@ -0,0 +1,46 @@
+from typing import Dict
+from database.models import TeamIndex
+from database.postgres.postgres_country import save_country
+
+def save_team(team_data: Dict, session, sport_id):
```
Avariq, line 5: all the same here

we should just pass a Team instance and save it. It's up to business layer to handle all the validation and calculation. We only care about saving to / reading from the database

database/postgres/save_data.py
```diff
@@ -0,0 +1,36 @@
+from database.session import SessionLocal
+from database.models import Sport
+from typing import Dict
+from database.postgres.postgres_team import save_teams
+from database.postgres.postgres_league import save_leagues
+from database.postgres.postgres_game import save_games
+
+def save_api_data(json_data: Dict, sport_name: str ) -> None:
```
Avariq, line 8: Same here

database/postgres/save_data.py
```diff
@@ -0,0 +1,36 @@
...
+            sport_id = sport.sport_id
+        else:
+            print(f"added new sport_id for {sport_name}.")
+            sport_entry = Sport(sport_name)
+            session.add(sport_entry)
+            session.commit()
+            sport_id = sport_entry.sport_id
+
+        entity = json_data.get("get")
+        if entity == "teams":
```
Avariq, line 22: We will have it replaced with models and think of a better way of doing this, but:

In general, when you expect some string to be present somewhere (doesn't matter where from exactly) you MUST specify that string somewhere depending on it's scope.

For example:

In your case you have your JSON data passed from API layer directly, which means that this string (let's say "teams") is somewhat global and might be used for comparison somewhere else in the code. If it's global (or expected to be used in diferent modules) we MUST create a separate class/module/whatever which will hold a set of predifined string to use.

For example:

`
class Strings:
    TEAMS = 'teams'
    LEAGUES = 'leagues'
    ...
`

database/postgres/save_data.py
```diff
@@ -0,0 +1,36 @@
...
+            save_teams(json_data, sport_id, session)
+            if sport_name == "mma" or sport_name == "formula-1":
+                save_leagues(json_data, sport_id, session)
+
+        elif entity == "leagues":
+            save_leagues(json_data, sport_id, session)
```
Avariq, line 28: We should create different DALs for each database entity. Let's talk about tomorrow

database/postgres/save_data.py
```diff
@@ -0,0 +1,169 @@
...
+            game_dal.save_games(game_dto_list)
+
+
+    except Exception as e:
+        session.rollback()
+        print("data was screwed up")
+        print(f"error processing data: {e}")
+
+
+def save_country_and_get_id(country_entry, country_dal: CountryDAL) -> int:
```
Avariq, line 130: `country_dal.create_country` method should already fetch the generated ID (since you are invoking `session.refresh(entity)`)

If that holds true (I might be wrong. But I do believe there should be a way of immediate Id fetching upon creation) - we can simplify quite a lot of logic and also have this method removed (`save_country_and_get_id`)

Avariq (approved):

PR #11, opened 2025-01-08
Feature/api route logger

AndrewKozovyi:
UJP-134


logger/logger.py
```diff
@@ -0,0 +1,42 @@
+import logging
+import functools
+import os
+from flask import request
+
+def get_logger(name, log_file, level=logging.INFO):
```
Avariq, line 6: make this a class "Logger"

we don't want "stray" functions in our files unless very necessary

api/routes/scripts.py
```diff
@@ -1,21 +1,30 @@
...
+    api_routes_logger.info(f"Generated cache key: {hashed_key} for method {request.method} on path {request.path}")
+    return f"{request.method}:{request.path}:{hashed_key}"
 
 
 def check_positive_param(param_name):
     param_value = request.args.get(param_name, type=int)
     if param_value is None or param_value <= 0:
+        api_routes_logger.warning(f"Invalid parameter value for {param_name}: {param_value}")
         return False
+    api_routes_logger.info(f"Valid parameter value for {param_name}: {param_value}")
```
Avariq, line 29: Sounds mor like a Debug line to me

api/routes/scripts.py
```diff
@@ -1,21 +1,30 @@
...
+    api_routes_logger.info(f"Cache key for sport type: {sport_type}")
+    return f"sport_stream_{sport_type}"
 
 
 def post_cache_key():
     json_data = request.get_json(silent=True) or {}
     json_string = json.dumps(json_data, sort_keys=True)
-    return f"{request.method}:{request.path}:{hashlib.md5(json_string.encode('utf-8')).hexdigest()}"
+    hashed_key = hashlib.md5(json_string.encode('utf-8')).hexdigest()
+    api_routes_logger.info(f"Generated cache key: {hashed_key} for method {request.method} on path {request.path}")
```
Avariq, line 20: sounds more like a Debug line to me

logger/logger.py
```diff
@@ -0,0 +1,42 @@
...
+    return logger
+
+
+def log_function_call(logger):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            try:
+                json_data = request.get_json(silent=True)
+                logger.info(f"Call of {func.__name__}() з args={args}, kwargs={kwargs}, json_data={json_data}")
```
Avariq, line 31: 50/50 Debug/Info

logger/logger.py
```diff
@@ -0,0 +1,42 @@
...
+
+                result = func(*args, **kwargs)
+
+                logger.info(f"Result of {func.__name__}() = {result}")
```
Avariq, line 35: 100% Debug. Or even better to have this removed.

We only log something that may be useful later and while I agree that function invokation and its results are indeed useful - if we log them all then all our log becomes bloody unreadable.

So it's a balance between having as much as possible without having it all together

api/routes/api_news.py
```diff
@@ -1,49 +1,63 @@
...
 @news_app.route('/recent', methods=['GET'])
 @cache.cached(timeout=60*60)
+@log_function_call(api_routes_logger)
 def get_recent_news_endpoint():
     try:
         recent_news = get_news_by_count(COUNT_NEWS)
         return recent_news
     except Exception as e:
+        api_routes_logger.error(f"Error in GET /: {str(e)}")
         get_error_response(e)
```
Avariq, line 32: BTW, I know it's not yours, but should it be `return get_error_response(e)`?

api/routes/scripts.py
```diff
@@ -1,21 +1,30 @@
...
+from logger.logger import get_logger
+
+api_routes_logger = get_logger("api_routes_loger", "api_routes.log")
+
 
 def get_cache_key():
-   sport_type = request.view_args.get('sport_type')
-   return f"sport_stream_{sport_type}"
+    sport_type = request.view_args.get('sport_type')
+    api_routes_logger.info(f"Cache key for sport type: {sport_type}")
```
Avariq, line 12: Debug here as well probably (may be wrong though. I don't know what it does)

service/api_logic/games_logic.py
```diff
@@ -27,6 +31,12 @@ def get_games_today(
             home_team.logo.label("home_team_logo"),
             away_team.name.label("away_team_name"),
             away_team.logo.label("away_team_logo"),
+            Games.score_home_team,
```
Avariq, line 34: crap, a lot of direct Database usage here too.

Don't think we have time to refactor everything again, so just remember that this is incorrect (watch Vlad,s PR. I wrote a memento there regarding this).

service/api_logic/games_logic.py
```diff
@@ -45,9 +55,13 @@ def get_games_today(
 
     offset, limit = pagination.get_pagination()
     if offset is not None and limit is not None:
+        api_logic_logger.info(f"Applying pagination: offset={offset}, limit={limit}")
```
Avariq, line 58: I don't think we need to log this

Keep in mind that YOU will be hostng this app. And you'll be getting thousands of different requests every hour (worst case estimation). I don't think you will need all this stuff in your log files wasting memory

service/api_logic/news_logic.py
```diff
@@ -6,27 +6,45 @@
...
+api_logic_logger = get_logger("api_logic_logger", "api_logic.log")
 
 session = SessionLocal()
 
+@log_function_call(api_logic_logger)
 def fetch_news(session, order_by: ClauseElement = None, limit: int = None, filters=None):
     query = session.query(News)
     if filters:
         query = query.filter(*filters)
+        api_logic_logger.info(f"Filters: {filters} applied")
```
Avariq, line 20: remove

service/api_logic/news_logic.py
```diff
@@ -6,27 +6,45 @@
...
+    else:
+        api_logic_logger.warning(f"Filters weren't applied. It returned {filters}")
+
     if order_by is not None:
         query = query.order_by(order_by)
+        api_logic_logger.info(f"Order: {order_by} applied")
```
Avariq, line 26: remove

service/api_logic/news_logic.py
```diff
@@ -6,27 +6,45 @@
...
+    else:
+        api_logic_logger.warning(f"Order wasn't applied. It returned {order_by}")
+
     if limit is not None:
         query = query.limit(limit)
+        api_logic_logger.info(f"Limit: {limit} applied")
```
Avariq, line 32: remove

service/api_logic/news_logic.py
```diff
@@ -6,27 +6,45 @@
...
+    else:
+        api_logic_logger.warning(f"Limit wasn't applied. It returned {limit}")
```
Avariq, line 34: remove

service/api_logic/news_logic.py
```diff
@@ -6,27 +6,45 @@
...
         query = query.filter(*filters)
+        api_logic_logger.info(f"Filters: {filters} applied")
+    else:
+        api_logic_logger.warning(f"Filters weren't applied. It returned {filters}")
+
     if order_by is not None:
         query = query.order_by(order_by)
+        api_logic_logger.info(f"Order: {order_by} applied")
+    else:
+        api_logic_logger.warning(f"Order wasn't applied. It returned {order_by}")
```
Avariq, line 28: remove

service/api_logic/news_logic.py
```diff
@@ -6,27 +6,45 @@
...
 session = SessionLocal()
 
+@log_function_call(api_logic_logger)
 def fetch_news(session, order_by: ClauseElement = None, limit: int = None, filters=None):
     query = session.query(News)
     if filters:
         query = query.filter(*filters)
+        api_logic_logger.info(f"Filters: {filters} applied")
+    else:
+        api_logic_logger.warning(f"Filters weren't applied. It returned {filters}")
```
Avariq, line 22: remove

api/routes/api_news.py
Iarynovskyi: It makes return in the middle of funk

service/api_logic/sports_logic.py
```diff
@@ -65,6 +74,9 @@ def search_leagues(filters_dto: dict, pagination: Pagination):
     offset, limit = pagination.get_pagination()
     if offset is not None and limit is not None:
         query = query.offset(offset).limit(limit)
+        api_logic_logger.info(f"Applying pagination: offset={offset}, limit={limit}")
```
Avariq, line 77: remove

service/api_logic/teams_logic.py
```diff
@@ -5,39 +5,53 @@
...
         "leagues": League,
     }
 
     query = apply_filters(query, filters_dto, model_aliases)
+    count = query.count()
 
     offset, limit = pagination.get_pagination()
     if offset is not None and limit is not None:
         query = query.offset(offset).limit(limit)
+        api_logic_logger.info(f"Applying pagination: offset={offset}, limit={limit}")
```
Avariq, line 44: remove

service/api_logic/games_logic.py
Iarynovskyi: you said then that someone would sit down and redo everything, so that's why this shit here ,'(

api/routes/scripts.py
```diff
@@ -1,21 +1,30 @@
-from flask import jsonify
+import logging
 from flask import request
 import hashlib
 import json
+from logger.logger import get_logger
+
+api_routes_logger = get_logger("api_routes_loger", "api_routes.log")
```
Avariq, line 7: Infinite log file? Waste of server memory.

In your new Logger class (which you will create of course) add some logic to create new log file when it's capacity threshold is exceeded (let's say 20Mb or something) and also remove old log files (we don't want to store a million of files).

Ask GPT about the best approach. Maybe some library already does everything you need

api/routes/scripts.py
Iarynovskyi: Func makes cache for post methods

logger/logger.py
```diff
@@ -0,0 +1,59 @@
...
+            logger.setLevel(self.level)
+
+        return logger
+
+    def log_function_call(self):
+        def decorator(func):
+            @functools.wraps(func)
+            def wrapper(*args, **kwargs):
+                try:
+                    self.logger.info(f"Call of {func.__name__}()  args={args}, kwargs={kwargs}")
```
Avariq, line 41: I think DEBUG or even TRACE would be better here

Avariq (approved):

PR #12, opened 2025-01-08
UJP-89 Feature/main page frontend

Iarynovskyi:
UJP-89


frontend/src/pages/leaguePage.js
```diff
@@ -1,77 +1,187 @@
...
-                        returnedTeams = league?.team
-                        setTeams(returnedTeams);
-                    }
-                })
-            })
-            .catch(error => {
-                toast.error(`:( Troubles With Teams Loading ${error}`);
-            });
-    }, []);
+        console.log("Sport ID:", sportId);
```
Avariq, line 82: seems like debug-only lines

probably best to have them removed

frontend/src/pages/leaguePage.js
```diff
@@ -1,77 +1,187 @@
...
+                        setLoading(false);
+                    }, 2000)
+            )
+    }, [loading]);
+
+
 
     return(
         <>
+            <section className={"leaguePage"}>
```
Avariq, line 113: Leave it as it is now, if you want, 'cause it's still readable (not too much html)

But maybe we could somehow make this html generic? So that we would be able to do something like: customModule(type=teams) /// customModule(type=leagues)  ... you get the idea. It would return the html we need with slight adjustments based on the type we require at the given moment. And then we would be able to reeuse it (mostly at least).

Of course, that is only possible if your leagues/teams/whatever are very similar. If they are not - consider everything written above to be BULLSHIT.

That pagination block just has some "generic" ring to it in my opinion

Avariq (approved):

PR #13, opened 2025-01-09
Finished refactor

IceStorman:


service/implementation/auto_request_api/sport_data_managers/teams_data_manager.py
```diff
@@ -0,0 +1,54 @@
+from typing import Optional
+
+from database.models import Sport
+from database.session import SessionLocal
```
Avariq, line 4: we shouldn't use database layer directly in BusinessLogic. I mean we can't expose Session or Database Models (tables) to our BusinessLayer since it violates SOLID. 

Data Access Layer should be separate and only expose its functionality through interfaces (public classes in our case. Some DALs. (like the ones Vika has made)).

service/implementation/auto_request_api/sport_data_managers/abstract_sport_data_manager.py
```diff
@@ -1,45 +1,51 @@
 import requests
 
-from dto.api_input import GamesDTO
+from dto.api_input import BaseDTO
 from service.implementation.auto_request_api.logic_auto_request import token_usage, api_key
 from database.azure_blob_storage.save_get_blob import blob_save_specific_api, get_all_blob_indexes_from_db, get_blob_data_for_all_sports
 from database.session import SessionLocal
```
Avariq, line 6: Same here. Exposed Session in Business Logic

service/implementation/auto_request_api/sport_data_managers/sport_consts.py
```diff
@@ -1,63 +1,69 @@
 from typing import Optional
 
-
-def get_team_statistics_url(sport_name: str, team_id: Optional[int], league_id: Optional[int]):
+def get_team_statistics_url(sport_id: str, team_id: Optional[int], league_id: Optional[int]):
     print("Getting team statistics url")
-    match sport_name:
+    match sport_id.lower():
+        case "handball":
```
Avariq, line 6: repeated strings OR strings that only have meaning when written correctly OR string that identify something ARE ALWAYS stored somewhere and the accessed.

get_team_statistics_url and get_team_statistics_index methods both have strings that should be identical, but still written separately. If some time in the future something changed (let's say "handball" had to be renamed to "TarasIsHappyToDoTheFrontendTaskAnytime") it would require somebody to literally go CTRL+F through the WHOLE project and rename each such string manually.

We want to prevent this, so please move all those common strings somewhere (new class, some enum, whatever you think is reasonable). 

NOTE (though obvious):

If our strings-containing class is expected to be used ONLY by BusinessLogic - we create it inside BusinessLogic layer (same goes for other layers, of course)
If, however, we intend to use those string in multiple places, then, in order to preserve our beloved SOLID, we would go and create new Layer/Project (let's call it "Common.Strings" for example). That layer/project will containt different classes with strings which can then be used wherever we want in our project (API/BL/DAL - don't care. SOLID stands).
For example one such class can be "SportsStrings" and it will contain something like: `volleyball_string = "volleyball"` and so on.

NOTE:

1. All the names for the classes I wrote should be reconsidered since I didn't put much thought to those.
2. Always ask ChatGPT for the best practices when it comes to naming classes, variables in them. I bet GPT knows exactly the best way of storing those strings of yours

service/implementation/auto_request_api/sport_data_managers/teams_data_manager.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class TeamsDataManager(AbstractSportDataManager):
+    _teams__sport_id: Optional[int]
+    _leagues__api_id: Optional[int]
+    _countries__api_id: Optional[int]
+    _page: Optional[int]
+    _per_page: Optional[int]
+
+    def __init__(self, data_manager):
```
Avariq, line 20: would be awesome if we could do something like: `self, data_manager: SomeSpecificDataManagerInterface`

just to make it more developer-friendly

service/implementation/auto_request_api/sport_data_managers/teams_data_manager.py
```diff
@@ -0,0 +1,54 @@
...
+            self._teams__sport_id = ix.sport_name
+        #self._teams__sport_id = data_manager.get("teams__sport_id")
+        self._leagues__api_id = data_manager.get("leagues__api_id")
+        self._host = get_host(self._teams__sport_id)
+
+        #self._sport_name = get_sport_name(self._teams__sport_id)
+
+        #self._countries__api_id = data_manager.get("countries__api_id")
+        self._page = data_manager.get("page")
+        self._per_page = data_manager.get("per_page")
```
Avariq, line 41: MINOR:

I would make the names slightly more expressive:

page_number
elements_per_page

something like that

tou always want your code to be as self-explanatory as you possibly can (no one wants to launch their shit two months later in order to debug some freaking line noone remebers ever being written in the first place just to check what it does)

service/implementation/auto_request_api/sport_data_managers/teams_data_manager.py
Avariq: I presume it is a dictionary here. If that is true then "data_manager" is not the best variable name for it (since it's just storeing data, not managing it)

service/api_logic/teams_logic.py
```diff
@@ -18,9 +16,11 @@ def get_teams(
     query = (
         session.query(TeamIndex)
          .join(League, TeamIndex.league == League.league_id)
-         .join(Country, TeamIndex.country == Country.country_id)
          .join(Sport, TeamIndex.sport_id == Sport.sport_id)
     )
+    a = query.all()
```
Avariq, line 21: wonderful variable naming (please rewrite immediately, wash your hands and go to church afterwards so that you can atone for this shit)

api/routes/api_teams.py
```diff
@@ -25,7 +24,8 @@ def get_teams_sport_endpoint():
         data = request.get_json()
         dto = TeamsLeagueDTO().load(data)
         pagination = Pagination(**data)
-        league_teams = get_teams(dto, pagination)
+        data_manager = TeamsDataManager(dto)
+        league_teams = data_manager.get_teams_data(pagination)
```
Avariq, line 28: MINOR:

it's better to separate "return" statement with a new line to make it stand out. Improves readability

service/implementation/auto_request_api/sport_data_managers/abstract_sport_data_manager.py
```diff
@@ -1,45 +1,51 @@
...
 
-    def main_request(host, name, url, blob_name):
+        query = (
+            SessionLocal().query(
+                Sport.sport_id,
+                Sport.sport_name
+            )
+            .filter(Sport.sport_id == new_data.get("sport_id"))
+        )
+        print(query)
```
Avariq, line 31: print() seems redundant

service/implementation/auto_request_api/sport_data_managers/sport_consts.py
```diff
@@ -1,63 +1,97 @@
...
+    football = "football"
+    nfl = "nfl"
+    mma = "mma"
+    formula_1 = "formula-1"
 
-def get_team_statistics_url(sport_name: str, team_id: Optional[int], league_id: Optional[int]):
-    print("Getting team statistics url")
-    match sport_name:
-        case "afl":
+def get_sport_type(sport_id: str) -> Optional[SportType]:
```
Avariq, line 17: Don't like the idea of having some specific methods in CONSTS file. Will have to discuss this

PR #16, opened 2025-01-27
UJP-95, UJP-99, UJP-97     login, sing-up, password-reset

Iarynovskyi:
UJP-95, UJP-99, UJP-97


service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+                </p>
+            </div>
+        </body>
+    </html>
+    """
+
+
+@handle_exceptions
+@api_logic_logger.log_function_call()
+class UserService:
```
Avariq, line 41: Should be a separate file for this

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
+from flask import current_app, url_for
+from flask_mail import Message
+from itsdangerous import URLSafeTimedSerializer
+from dto.api_output import OutPutUser
+from database.models import User
+from exept.handle_exeptions import handle_exceptions
+import bcrypt
+from logger.logger import Logger
+
+api_logic_logger = Logger("api_logic_logger", "api_logic_logger.log")
```
Avariq, line 10: We are in BusinessLayer, so I don't see any reason which would justify this being named "api_logic_logger". Just "log" or "logger" will suffice

Avariq, line 10: Also, this logger is used ONLY within the UserService, so we should move it inside the UserService class.

BUT be careful: Your UserService is registered as Scoped in your DI Container, which will create new Instance of this class upon every request from API. We don't want a separate logger for each UserService instance right?

Consider adding logger to the container as well, and just passing it inside UserService as a **Singleton** (we only need one instance of logger, because its job is virtually just printing text is is provided with, it has no state to maintain)

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+
+def message_to_user_gmail(user: User, reset_url):
```
Avariq, line 12: While I don't mind you hardcoding the html like this there are a couple of things to consider:

1) It looks like a resource to me. Something like our enums we have for sports and so on. I would better have it stored separately with our other resources. (Нагадайте про це на мітингу, будь ласка, якщо я забуду)

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+            return OutPutUser().dump(None)
+
+        token = self.get_reset_token(existing_user.email)
+
+        return self.send_reset_email(existing_user, token)
+
+    def send_reset_email(self, user: User, token: str):
+        try:
+            reset_url = url_for('login_app.reset_password', token=token, _external=True)
+            print(message_to_user_gmail(user, reset_url))
```
Avariq, line 72: we don't use prints anymore.

if it is something important - use logger, if not - remove

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+    </html>
+    """
+
+
+@handle_exceptions
+@api_logic_logger.log_function_call()
+class UserService:
+    def __init__(self, user_dal):
+        self.user_dal = user_dal
+        self.s = URLSafeTimedSerializer(current_app.secret_key)
```
Avariq, line 44: rename the property

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            reset_url = url_for('login_app.reset_password', token=token, _external=True)
+            print(message_to_user_gmail(user, reset_url))
+            msg = Message(
+                'Password Reset Request',
+                sender=current_app.config['MAIL_USERNAME'],
+                recipients=[user.email],
+                html=message_to_user_gmail(user, reset_url)
+            )
+            #msg.body = message_to_user_gmail(user, reset_url)
```
Avariq, line 79: remove if not used

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+    def reset_user_password(self, email, new_password: str) -> None:
+        user = self.user_dal.get_user_by_email(email)
+        self.user_dal.update_user_password(user, new_password)
+
+    def confirm_token(self, token: str, expiration=3600):
+        try:
+            email = self.s.loads(token, salt="email-confirm", max_age=expiration)
+        except:
+            return False
+        user = self.user_dal.get_user_by_email(email)
```
Avariq, line 97: add empty line for readability

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+    def get_reset_token(self, email) -> str:
+        return self.s.dumps(email, salt="email-confirm")
+
+    def reset_user_password(self, email, new_password: str) -> None:
+        user = self.user_dal.get_user_by_email(email)
+        self.user_dal.update_user_password(user, new_password)
+
+    def confirm_token(self, token: str, expiration=3600):
+        try:
+            email = self.s.loads(token, salt="email-confirm", max_age=expiration)
```
Avariq, line 94: Нагадайте розказати вам про salting на мітингу. Воно трохи не так юзається

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+        user = self.user_dal.get_user_by_email(email)
+        return OutPutUser().dump(user)
+
+    def log_in(self, email_or_username: str, password: str):
+
+        user = self.user_dal.get_user_by_email_or_username_and_password(email_or_username, password)
+
+        if user:
+            token = self.generate_auth_token(user)
+            return {"message": "Успішний вхід", "token": token}
```
Avariq, line 106: We either use full English everywhere or full Ukraininan. Since we decided to go fith former - please change accordingly

service/api_logic/preferences_logic.py
```diff
@@ -0,0 +1,31 @@
+from exept.handle_exeptions import handle_exceptions
+from logger.logger import Logger
+from dto.api_output import OutputPreferences
+
+api_logic_logger = Logger("api_logic_logger", "api_logic_logger.log")
+
+
+@handle_exceptions
+@api_logic_logger.log_function_call()
+class PreferencesService:
```
Avariq, line 10: Should be reimagined. Discussed with Dima in private

api/routes/api_login.py
```diff
@@ -0,0 +1,117 @@
+from flask import Blueprint, jsonify, request
+from dto.api_input import InputUserDTO, ResetPasswordDTO, NewPasswordDTO, InputUserLoginDTO
+from exept.exeptions import DatabaseConnectionError
+from exept.handle_exeptions import get_error_response, get_error_reset_password, get_good_reset_password
+from logger.logger import Logger
+from dependency_injector.wiring import inject, Provide
+from service.api_logic.login_logic import UserService
+from api.container.container import Container
+from flask_jwt_extended import JWTManager,create_access_token, set_access_cookies, unset_jwt_cookies
```
Avariq, line 9: JWTManager is imported but never used

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+
+        new_user = User(email=email_front, username=username_front, password_hash=hashed_password.decode('utf-8'))
+        self.user_dal.create_user(new_user)
+
+        return OutPutUser().dump(new_user)
+
+    def request_password_reset(self, email: str):
+        existing_user = self.user_dal.get_user_by_email(email)
+        if not existing_user:
+            return OutPutUser().dump(None)
```
Avariq, line 63: Rename to OutputUser (output is a single word, not a combination of two. A little misleading)

service/api_logic/login_logic.py
```diff
@@ -0,0 +1,111 @@
...
+        user = self.user_dal.get_user_by_email_or_username_and_password(email_or_username, password)
+
+        if user:
+            token = self.generate_auth_token(user)
+            return {"message": "Успішний вхід", "token": token}
+
+        return {"error": "Невірні облікові дані"}
+
+    def generate_auth_token(self, user):
+        return self.s.dumps(user.email, salt="user-auth-token")
```
Avariq, line 111: Salting

We should move all the JWT-related logic to some other class/manager/service/whatever

Нагадайте на мітингу

api/routes/api_login.py
```diff
@@ -0,0 +1,117 @@
...
+
+@login_app.route('/sign_up', methods=['POST'])
+@inject
+@api_routes_logger.log_function_call()
+def create_account_endpoint(service: UserService = Provide[Container.user_service]):
+    try:
+        data = request.get_json()
+        dto = InputUserDTO().load(data)
+
+        result = service.create_user(dto["email"], dto["username"], dto["password_hash"])
```
Avariq, line 33: can't we just do `dto.email` for example?

service/api_logic/login_logic.py
Avariq: Не бачу ніде ніяких приватних ключів для підпису токена, поговоримо про то на мітингу (нагадаєте)

api/routes/__init__.py
```diff
@@ -1,17 +1,57 @@
...
+    app.secret_key = "SECRET_KEY"
+    app.config['MAIL_SERVER'] = 'smtp.gmail.com'
+    app.config['MAIL_PORT'] = 587
+    app.config['MAIL_USE_TLS'] = True
+    app.config['MAIL_USE_SSL'] = False
+    app.config['MAIL_USERNAME'] = 'q.sport.news@gmail.com'
+    app.config['MAIL_PASSWORD'] = 'jjrc siyp trnq tzcp'
+    mail.init_app(app)
+    
+    app.config['JWT_SECRET_KEY'] = 'your_secret_key'
```
Avariq, line 49: Secret key (generate)

api/routes/api_login.py
```diff
@@ -0,0 +1,117 @@
...
+        return get_error_response(e)
+
+@login_app.route('/login', methods=['POST'])
+@inject
+@api_routes_logger.log_function_call()
+def log_in(service: UserService = Provide[Container.user_service]):
+    try:
+        data = request.get_json()
+
+        dto = InputUserLoginDTO().load(data)
```
Avariq, line 91: Розібратися з автовалідацією

exept/handle_exeptions.py
```diff
@@ -27,6 +27,13 @@ def get_error_response(e):
     response = {"error in service": str(e)}
     return jsonify(response)
 
+def get_error_reset_password():
+    response = {"message": "Invalid or expired token"}
+    return jsonify(response), 400
+
+def get_good_reset_password():
```
Avariq, line 34: good response in except

exept/handle_exeptions.py
```diff
@@ -27,6 +27,13 @@ def get_error_response(e):
...
+    response = {"message": "Success password rest"}
```
Avariq, line 35: винесіть в модельку

service/api_logic/user_logic.py
```diff
@@ -0,0 +1,96 @@
...
+    def send_reset_email(self, user: User, token: str):
+        reset_url = url_for('login_app.reset_password', token=token, _external=True)
+        msg = Message(
+            'Password Reset Request',
+            sender=current_app.config['MAIL_USERNAME'],
+            recipients=[user.email],
+            html=self.message_to_user_gmail(user, reset_url)
+        )
+        current_app.extensions['mail'].send(msg)
+        return {"msg": "Success"}
```
Avariq, line 62: We either return nothing (preferred) of we make a dto for common results like this

service/api_logic/user_logic.py
```diff
@@ -0,0 +1,96 @@
...
+            return False
+        user = self.user_dal.get_user_by_email(email)
+
+        return OutPutUser().dump(user)
+
+    def log_in(self, email_or_username: str, password: str):
+
+        user = self.user_dal.get_user_by_email_or_username_and_password(email_or_username, password)
+
+        if user:
```
Avariq, line 89: if not user:
      we throw exception...

return success


less code - better readability - better SOLID

service/api_logic/user_logic.py
```diff
@@ -0,0 +1,96 @@
...
+            token = self.generate_auth_token(user)
+            return {"message": "Success", "token": token}
+
+        return {"error": "Incorrect data"} #Тарас сказав
+
+    def generate_auth_token(self, user):
+        return self.serializer.dumps(user.email, salt="user-auth-token")
```
Avariq, line 96: don't forget to fix the salt

service/api_logic/user_logic.py
```diff
@@ -0,0 +1,96 @@
...
+from dto.api_output import OutPutUser, OutputPreferences
+from database.models import User
+from exept.handle_exeptions import handle_exceptions
+import bcrypt
+from logger.logger import Logger
+from jinja2 import Environment, FileSystemLoader
+import os
+from flask_jwt_extended import create_access_token
+
+api_logic_logger = Logger("api_logic_logger", "api_logic_logger.log")
```
Avariq, line 13: Don't like the idea of having logger (BTW - change the name to either just "log" or "logger") here. It cannot exist without UserService, right? Then it should be within the said service (Singleton)

service/api_logic/templates/reset_password_email.html
```diff
@@ -0,0 +1,27 @@
...
+</head>
+    <body style="font-family: Arial, sans-serif; line-height: 1.6; color: #333; background-color: #f9f9f9; padding: 20px;">
+        <div style="max-width: 600px; margin: 0 auto; background: white; border: 1px solid #ddd; border-radius: 10px; padding: 20px;">
+            <h2 style="color: #4CAF50; text-align: center;">Password Reset Request</h2>
+            <p>Dear <strong>{{ username }}</strong>,</p>
+            <p>You have requested to reset your password. To do so, click the button below:</p>
+            <div style="text-align: center; margin: 20px 0;">
+                <a href="{{ reset_url }}" style="display: inline-block; background-color: #4CAF50; color: white; text-decoration: none; padding: 10px 20px; border-radius: 5px; font-size: 16px;">Reset Your Password</a>
+            </div>
+            <p>If the button above doesn't work, copy and paste the following link into your browser:</p>
```
Avariq, line 15:
```suggestion
            <p>If the button above does not work, follow the link below:</p>
```

"doesn't" is informal. we should be using formal English when contacting our users

exept/handle_exeptions.py
```diff
@@ -27,6 +27,10 @@ def get_error_response(e):
     response = {"error in service": str(e)}
     return jsonify(response)
 
+def get_error_reset_password():
```
Avariq, line 30: Remove this method. Make exception instead

dto/api_input.py
```diff
@@ -56,3 +62,30 @@ class GamesDTO(BaseDTO):
...
+    email_or_username = fields.Str(required=True)
+    password_hash = fields.Str(required=True)
+
+
+class InputUserPreferencesDTO(BaseDTO):
+    preferences = fields.List(fields.Int, required=True)
+    user_id = fields.Int(required=False, missing=None)
+
+
+class InputGetUserPreferencesDTO(BaseDTO):
```
Avariq, line 90: InputGetUserPreferencesDTO and InputUserPreferencesDTO what's the difference between the two and why do we need both? (I also suggest renaming them to some more specific names. "Get" part is not adding nuch of a context IMO)

dto/api_input.py
```diff
@@ -56,3 +62,30 @@ class GamesDTO(BaseDTO):
...
+    email = fields.Str(required=True)
+
+
+class NewPasswordDTO(BaseDTO):
+    new_password = fields.Str(required=True)
+    email = fields.Str(required=True)
+
+
+class InputUserLoginDTO(BaseDTO):
+    email_or_username = fields.Str(required=True)
```
Avariq, line 81: Either email or username.

Those two fields have to be validated separately (e.g. we need to check whether the email is in correct format as well as check whether username is correct, has no whitespaces and so on)

Write a Marshmellow-compatible validator for that and use it here (we need either a correct username or correct email. One of the two, but not NONE. Well, I bet you'll figure it out. Good luck)

database/postgres/dal/user.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def get_user_by_email_or_username_and_password(self, email_or_username, password):
+        user = self.session.query(User).filter(
+            or_(
+                User.email == email_or_username,
+                User.username == email_or_username
+            )
+        ).first()
+
+        if user and bcrypt.checkpw(password.encode('utf-8'), user.password_hash.encode('utf-8')):
```
Avariq, line 39: This should be done in BusinessLayer. We don't do logic inside DAls.

Just return the User model and it will contain the password hash you need (remeber: It has to be a DTO, not an actual sqlachemy User model)

Avariq: If you move this logic away you actually may remove this method altogether, since you already have one just like this

database/postgres/dal/user.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def get_user_by_email_or_username(self, email, username):
+        return self.session.query(User).filter(
+            or_(
+                User.email == email,
+                User.username == username
+            )
+        ).first()
+
+    def get_user_by_email(self, email: str) -> User:
```
Avariq, line 17: Do we need `get_user_by_email` if we already have `get_user_by_email_or_username`?

Also it doesn't add up to "email_or_username" I saw in a DTO model somehere on API level. Let's stick to one approach only (either separate username and email or joined email_or_username). Change the logic accordingly

api/routes/api_login.py
```diff
@@ -0,0 +1,103 @@
...
+        data = request.get_json()
+        dto = InputUserLoginDTO().load(data)
+        user = service.log_in(dto.email_or_username, dto.password_hash)
+
+        if not user:
+            return {"error": "Invalid email/username or password"}, 401
+
+        access_token = create_access_token(identity={'id': user.id, 'email': user.email})
+
+        response = {"message": "You successfully loged in!", "user": {"id": user.id, "email": user.email}}
```
Avariq, line 91:
```suggestion
        response = {"message": "You successfully logged in!", "user": {"id": user.id, "email": user.email}}
```
PyCharm has a spellcheck in it. Free of charge :)

Avariq, line 91: `"user": {"id": user.id, "email": user.email}` -> DTO for this. Keep in mind - we don't want a million of DTOs, but we certainly don't want plain text/dictionary being returned like this either. Find a way to satisfy both of the conditions

api/routes/api_login.py
```diff
@@ -0,0 +1,103 @@
...
+        return get_error_response(e)
+
+@login_app.route('/login', methods=['POST'])
+@inject
+@api_routes_logger.log_function_call()
+def log_in(service: UserService = Provide[Container.user_service]):
+    try:
+        data = request.get_json()
+        dto = InputUserLoginDTO().load(data)
+        user = service.log_in(dto.email_or_username, dto.password_hash)
```
Avariq, line 84: it's better if `service.log_in()` just throws relevant exception instead of returning NULL/None values. Generally it's a bad practice if your method returns NULL/None without a good reason (which this one is not)

dto/api_input.py
```diff
@@ -16,6 +17,11 @@ def clean_letter(self, data, **kwargs):
             data['letter'] = ' '.join(data['letter'].split()).lower()
         return data
 
+    @post_load
+    def make_object(self, data, **kwargs):
```
Avariq, line 21: What is this for? (just asking)

Iarynovskyi: This IS for (  dto.something  )

dto/api_input.py
AndrewKozovyi: Validation is complete.
email or username is used as 1 varaiable, as we want our user to enter either their username or email, sowe watn this as 1 filed. Do you mean that we use dto with email and name as different fields, but have both those paramets equal to 1 value?

PR #18, opened 2025-02-03
UJP-9 user preferences

Iarynovskyi:
problems with alembic fixed here


api/routes/api_user_preferences.py
```diff
@@ -0,0 +1,78 @@
...
+from exept.exeptions import DatabaseConnectionError, CustomQSportException
+from exept.handle_exeptions import get_custom_error_response, handle_exceptions
+from logger.logger import Logger
+from dependency_injector.wiring import inject, Provide
+from service.api_logic.user_logic import UserService
+from api.container.container import Container
+
+
+
+api_routes_logger = Logger("api_routes_logger", "api_routes_logger.log")
```
Avariq, line 12: just "logger"

```suggestion
logger = Logger("api_routes_logger", "api_routes_logger.log")
```

database/postgres/dal/preferences.py
```diff
@@ -0,0 +1,87 @@
...
+            user_id_field=user_id_field,
+            type_id_field=type_id_field,
+            related_name=related_name,
+            related_logo=related_logo,
+            related_id=related_id
+        )
+
+
+    def get_all_sport_preferences(self):
+        return get_all_sports()
```
Avariq, line 34: your DAL should not depend on BL layer.

Basically, you can't use neither API nor BL inside your database project

exept/exeptions.py
```diff
@@ -83,4 +83,28 @@ class UserAlreadyExistError(CustomQSportException):
...
+
+class IncorrectPreferencesError(CustomQSportException):
+    status_code = 400
+
+    def __init__(self):
+        message = f"No valid preferences"
+        super().__init__(message)
+
+
+class NotUserIdOrPreferencesError(CustomQSportException):
```
Avariq, line 97: NotUserId... naming is very odd

database/postgres/dal/preferences.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+    def add_user_preferences(self, type_dto, dto, valid_preferences):
+        tables_and_cols_dto = self.getattr_tables_and_columns_by_type(type_dto)
+
+        for pref in valid_preferences:
```
Avariq, line 40: Can't we use single query for that instead?

Avariq: And BTW this does not belong in DAL, because it performs logic. DAL should only dance around CRUD operations at their simplest, so move all logic to BL

dto/api_input.py
```diff
@@ -121,14 +123,64 @@ class InputUserLoginDTO(BaseDTO):
...
+        self.main_table = "UserPreference"
+        self.related_table = "Sport"
+        self.user_id_field = "users_id"
+        self.type_id_field = "sports_id"
+        self.related_name = "sport_name"
+        self.related_logo = "sport_img"
+        self.related_id = "sport_id"
+
+
+class TeamPreferenceDTO:
```
Avariq, line 152: It's not a DTO if it just contains some constants

dto/api_output.py
```diff
@@ -44,11 +44,18 @@ class OutputUser(Schema):
     username = fields.Str(required=True)
     email = fields.Str(required=True)
 
-class OutputPreferences(Schema):
+class OutputSportPreferences(Schema):
```
Avariq, line 47: Seems like `OutputSportPreferences` and `OutputTeamPreferences` can be put into a single DTO

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+        return response_json
+
+
+    def add_preferences(self, dto: UpdateUserPreferencesDTO):
+        if not dto.user_id or not isinstance(dto.preferences, list):
+            raise NotUserIdOrPreferencesError()
+
+        new_dto_by_type_of_preference = self.dto_for_type_of_preference(dto)
+
+        existing_sports = self._preferences_dal.get_all_sport_preference_indexes()
```
Avariq, line 147: We should cache this response if not already. I believe we don't expect the response to change often, right?

Maybe SQLAlchemy supports caching it responses. If not, we should probably store this response somewhere ourselves

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+        if dto.type == SPORT_TYPE:
+            valid_preferences = set([sport_id for sport_id in dto.preferences if sport_id in existing_sports])
+
+        if dto.type == TEAM_TYPE:
+            valid_preferences = set([sport_id for sport_id in dto.preferences])
```
Avariq, line 152: don't we need `if sport_id in existing_sports` here as well?

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+
+        if not valid_preferences:
+            raise IncorrectPreferencesError()
+
+        self._preferences_dal.delete_all_user_preferences(new_dto_by_type_of_preference, dto)
+        self._preferences_dal.add_user_preferences(new_dto_by_type_of_preference, dto, valid_preferences)
```
Avariq, line 158: What if new preferences differ just by one preference from a previous set? Will we still remove everything and add them from scratch?

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+
+        return CommonResponse().to_dict()
+
+
+    def get_user_preferences(self, dto: UpdateUserPreferencesDTO):
+        new_dto_by_type_of_preference = self.dto_for_type_of_preference(dto)
+
+        prefs = self._preferences_dal.get_user_preferences(new_dto_by_type_of_preference, dto)
+        if dto.type == SPORT_TYPE:
+            shema = OutputSportPreferences(many=True).dump(prefs)
```
Avariq, line 168: why `schema` all of a sudden?

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+        if dto.type == TEAM_TYPE:
+            valid_preferences = set([sport_id for sport_id in dto.preferences])
+
+        if not valid_preferences:
+            raise IncorrectPreferencesError()
+
+        self._preferences_dal.delete_all_user_preferences(new_dto_by_type_of_preference, dto)
+        self._preferences_dal.add_user_preferences(new_dto_by_type_of_preference, dto, valid_preferences)
+
+        return CommonResponse().to_dict()
```
Avariq, line 160: If you don't actually need any response from the BusinessLayer - don't add it in the first place.

Yes, currently your BL is only ever used by API layer, but what if I were to use this from another service? Would some "success" message even be helpful to me?

I'd go with void function (returns None) if I were you

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+        if dto.type == SPORT_TYPE:
+            shema = OutputSportPreferences(many=True).dump(prefs)
+            return shema
+
+        if dto.type == TEAM_TYPE:
+            shema = OutputTeamPreferences(many=True).dump(prefs)
+            return shema
+
+
+    def get_all_preferences(self):
```
Avariq, line 176: `get_all_preferences` method ONLY returns sport prefs. That is misleading.

service/api_logic/user_logic.py
```diff
@@ -128,4 +135,64 @@ async def create_access_token_response(self, user):
...
+            raise NotUserIdOrPreferencesError()
+
+        new_dto_by_type_of_preference = self.dto_for_type_of_preference(dto)
+
+        self._preferences_dal.delete_user_preferences(new_dto_by_type_of_preference, dto)
+
+        return CommonResponse().to_dict()
+
+
+    def dto_for_type_of_preference(self, dto):
```
Avariq, line 191: can be static

dto/api_input.py
```diff
@@ -121,14 +123,64 @@ class InputUserLoginDTO(BaseDTO):
...
+    def __init__(self):
```
Avariq, line 153: We should not expose our database structure on API level. I doubt our API cares where it gets it data from from table A or Table B. That is for BL and DAl to play around with

dto/api_input.py
```diff
@@ -121,14 +123,64 @@ class InputUserLoginDTO(BaseDTO):
...
+        self.main_table = "UserClubPreferences"
+        self.related_table = "TeamIndex"
+        self.user_id_field = "users_id"
+        self.type_id_field = "preferences"
+        self.related_name = "name"
+        self.related_logo = "logo"
+        self.related_id = "team_index_id"
+
+
+class TablesAndColumnsDTO:
```
Avariq, line 163: Firstly - it's only used for prefs and cannot possible be used for anything else, therefore your naming is incorrect.

Secondly, this is a very tricky approach which may backfire some time in the future. I am not gonna make you redesign this but I suggest you return to this task some time in the future and try to solve it again in a better way. Maybe after this project is done and presented.

dto/api_input.py
```diff
@@ -121,14 +123,64 @@ class InputUserLoginDTO(BaseDTO):
...
     id = fields.Str(required=False, missing=None)
     verified_email = fields.Bool(required=False, missing=None)
-    picture = fields.Str(required=False, missing=None)
\ No newline at end of file
+    picture = fields.Str(required=False, missing=None)
+
+
+class SportPreferenceDTO:
+    def __init__(self):
+        self.main_table = "UserPreference"
```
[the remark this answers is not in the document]
Iarynovskyi: Will I run a new migration to change one letter ?...

dto/api_output.py
Iarynovskyi: Seems, but I tried to make a universal method for sports and teams and they have
different fields names in the db tables

service/api_logic/user_logic.py
Iarynovskyi: no, because we have thousands of commands and I don't want to check each one 
to see if we have that one in db.  

and here must be team_id... 
i missed this naming

service/api_logic/user_logic.py
Iarynovskyi: The API is breaking for some reason if i return None for example, idk why

database/postgres/dal/preferences.py
```diff
@@ -0,0 +1,93 @@
...
+            .filter(tables_and_cols_dto.user_id_field == dto.user_id)
+            .all()
+        )
+        return user_prefs
+
+
+    def delete_user_preferences(self, type_dto, dto):
+        tables_and_cols_dto = self.getattr_tables_and_columns_by_type(type_dto)
+
+        for pref in dto.preferences:
```
Avariq, line 53: We can remove them all with one query

service/api_logic/user_logic.py
Avariq: Will discuss on the meeting

service/api_logic/user_logic.py
```diff
@@ -138,6 +146,57 @@ async def create_access_token_response(self, user):
...
+        prefs = self._preferences_dal.get_user_preferences(new_dto_by_type_of_preference, dto)
+        if dto.type == SPORT_TYPE:
+            return OutputSportPreferences(many=True).dump(prefs)
+
+        if dto.type == TEAM_TYPE:
+            return  OutputTeamPreferences(many=True).dump(prefs)
+
+
+    def get_all_sport_preferences(self):
+        return self._sport_dal.get_all_sports()
```
Avariq, line 179: We just return all the sports we have. I don't see how "preferences" in the name could be connected to that

api/routes/api_user_preferences.py
```diff
@@ -0,0 +1,78 @@
...
+    except CustomQSportException as e:
+        logger.error(f"Error in POST /: {str(e)}")
+        return get_custom_error_response(e)
+
+
+@preferences_app.route('/all', methods=['GET'])
+@inject
+@handle_exceptions
+@logger.log_function_call()
+def get_all_sport_preferences_endpoint(service: UserService = Provide[Container.user_service]):
```
Avariq, line 70: Do we even need this endpoint to begin with?

It certainly does not belong in "user_preferences" controller

dto/api_output.py
Avariq: Well, you can map them on BL layer, or just rename the DB table columns.

Why would we want to make our job harder in the future if it could be simplified right away

database/postgres/dal/sport.py
```diff
@@ -2,6 +2,8 @@
 from database.models import Sport
 from database.postgres.dto import SportDTO
 from typing import Optional
+from dto.api_output import SportsOutput
```
Avariq, line 5: DB should not depend on API or BL

PR #19, opened 2025-02-04
UJP-159 Google Auth pattern Srategy applied for user Log In

AndrewKozovyi:
UJP-159


api/routes/api_login.py
```diff
@@ -95,8 +96,21 @@ def reset_password(token, service: UserService = Provide[Container.user_service]
 async def log_in(service: UserService = Provide[Container.user_service]):
     try:
         data = request.get_json()
-        dto = InputUserLoginDTO().load(data)
-        user = await service.log_in(dto.email_or_username, dto.password_hash)
+        dto = T
+        status = ""
+
+        try:
+            dto = InputUserByGoogleDTO().load(data) or dto
```
Avariq, line 103: API endpoint (by following RES specifications) has to be unifrm -> accept and return predefined data.

Basically -> We always accept ONE dto and always return ONE dto.

Therefore, I suggest creating one DTO here for login which would be something like:

<google required fields>? (nullable)
<facebook required fields>? (nulable)
<our fields (login/email/password/whatever the default auth requires)>? (nullable)
auth_provider -> string -> MANDATORY (we have to check that value later against some predefined set of values, like Enum, to ensure we only process auth methods we expect). For example: GoogleAuth, FacebookAuth, SimpleAuth

You can write validation for the fileds manually (somewhere in business layer, or your strategy classes) or you can write custom validator for marshmello (alittle harder. It's up to you. Both options are fine)

api/routes/api_login.py
```diff
@@ -95,8 +96,21 @@ def reset_password(token, service: UserService = Provide[Container.user_service]
...
+            status = "google"
```
Avariq, line 104: Hardcoded string which will have to be moved to some predefined space (like Enum)

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+from dto.api_input import InputUserLoginDTO, InputUserByGoogleDTO
+from dto.api_output import OutputLogin
+from exept.exeptions import IncorrectUsernameOrEmailError, IncorrectPasswordError
+from typing import Generic, TypeVar
+
+from logger.logger import Logger
+
+T = TypeVar("T")
+
+class AuthStrategy(ABC, Generic[T]):
```
Avariq, line 19: Even though it's a strategy indeed - do not name it like that. The name must be self-explanatory and brief.

AuthHandler? AuthHelper? 

Think of some better naming for this

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+class AuthContext:
+    def __init__(self, strategy: AuthStrategy):
+        self._strategy = strategy
+
+    async def execute_login(self, credentials: T):
+        xuy = await self._strategy.authenticate(credentials)
+        return xuy
+
+
+class SimpleAuthStrategy(AuthStrategy[InputUserLoginDTO]):
```
Avariq, line 39: Better naming as well

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+
+        token = await self.__generate_auth_token(user)
+
+        return OutputLogin(email=user.email, token=token, id=user.user_id)
+
+    async def __generate_auth_token(self, user):
+            return self._serializer.dumps(user.email, salt = "user-auth-token")
+
+
+class GoogleAuthStrategy(AuthStrategy[InputUserByGoogleDTO]):
```
Avariq, line 59: Better naming as well

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+    def authenticate(self, credentials: T):
+        pass
+
+
+class AuthContext:
+    def __init__(self, strategy: AuthStrategy):
+        self._strategy = strategy
+
+    async def execute_login(self, credentials: T):
+        xuy = await self._strategy.authenticate(credentials)
```
Avariq: Shall I start punishing you for such variables?

While I do find it to be peak comedy - let's be as professional here, as we can, so that you all can add this project to your CVs later :)

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+        self._serializer = serializer
+
+    @abstractmethod
+    def authenticate(self, credentials: T):
+        pass
+
+
+class AuthContext:
+    def __init__(self, strategy: AuthStrategy):
+        self._strategy = strategy
```
Avariq, line 32: Naming

service/api_logic/auth_strategy.py
AndrewKozovyi: I've changed it before u started reviewing (name was like that, because I was mad, nothing worked:( )

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+from exept.exeptions import IncorrectUsernameOrEmailError, IncorrectPasswordError
+from typing import Generic, TypeVar
+
+from logger.logger import Logger
+
+T = TypeVar("T")
+
+class AuthStrategy(ABC, Generic[T]):
+    def __init__(self, user_dal, serializer):
+        self._user_dal = user_dal
```
Avariq, line 21: Your "strategies" are but a single method which acceps and processes some data. It does not store it in any way within itself, which makes it stateless. It also cannot exist without userservice and userdal, so it has strong dependencies.

That being said: It's more of a "sub-class", if you will, an extension to your userservice. That means that we don't need to store userdal and userservice each time. We already have both of those instances in our userservice. Let's just reuse them.

Same goes for logger. This strategy is a part of our userservice so it should be logged along with it with the same logger instance.

That being said: just pass userdal, userservice and logger from userservice as function params

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, strategy: AuthStrategy):
+        self._strategy = strategy
+
+    async def execute_login(self, credentials: T):
+        xuy = await self._strategy.authenticate(credentials)
+        return xuy
+
+
+class SimpleAuthStrategy(AuthStrategy[InputUserLoginDTO]):
+    async def authenticate(self, credentials: InputUserLoginDTO):
```
Avariq, line 40: When you change the DTO on API level to be general, you won't need `InputUserLoginDTO` and `InputUserByGoogleDTO` since they will basically merge

Avariq, line 40: pass userdal, userservice and logger here alongside credentials

service/api_logic/auth_strategy.py
AndrewKozovyi: Okay, I have better vision of how I can implement it now, thanks!

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+        user = self._user_dal.get_user_by_email_or_username(credentials.email_or_username, credentials.email_or_username)
```
Avariq, line 41: We can use userservice logic here, instead of userdal

service/api_logic/auth_strategy.py
Avariq: UPD: 
We don't need userdal whatsoever

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+        user = self._user_dal.get_user_by_email_or_username(user_info.email)
+        if not user:
+            user = User(email=user_info.email, username=user_info.email.split('@')[0])
+            self._user_dal.create_user(user)
+
+        token = self.__generate_auth_token(user)
+
+        return OutputLogin(email=user.email, token=token, id=user.user_id)
+
+    async def __generate_auth_token(self, user):
```
Avariq, line 88: This method seems to be duplicated.

AndrewKozovyi: That the thing I wrote about in telegram, I thought it was a bit (a lot) like copy-paste, but when u told about using userService, I think I will be able to use it from there

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+    async def authenticate(self, credentials: InputUserLoginDTO):
+        user = self._user_dal.get_user_by_email_or_username(credentials.email_or_username, credentials.email_or_username)
+
+        if not user:
+            self._logger.warning("User does not exist")
+            raise IncorrectUsernameOrEmailError()
+
+        if not bcrypt.checkpw(credentials.password_hash.encode('utf-8'), user.password_hash.encode('utf-8')):
+            self._logger.warning("Passwords do not match")
+            raise IncorrectPasswordError()
```
Avariq, line 49: I would do something about this exception, Don't like the idea of letting anybody know, that Username was correct, but passwrd wasn't. Let's just always say "Incorrect credentials" and be done with it

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,89 @@
...
+            current_app.config['USER_INFO_URL'],
+            headers={'Authorization': f'Bearer {client.token["access_token"]}'}
+        )
+
+        user_info = InputUserByGoogleDTO().load(user_info_response.json())
+
+        user = self._user_dal.get_user_by_email_or_username(user_info.email)
+        if not user:
+            user = User(email=user_info.email, username=user_info.email.split('@')[0])
+            self._user_dal.create_user(user)
```
Avariq, line 82: userservice instead

service/api_logic/user_logic.py
```diff
@@ -92,34 +103,16 @@ def confirm_token(self, token: str, expiration=3600):
...
-        if not bcrypt.checkpw(password.encode('utf-8'), user.password_hash.encode('utf-8')):
-            self._logger.warning("Passwords do not match")
-            raise IncorrectPasswordError()
-
-        token = await self.__generate_auth_token(user)
-
-        return OutputLogin(email = user.email, token = token, id = user.user_id)
+    async def log_in(self, method: str, credentials: Union[InputUserLoginDTO, InputUserByGoogleDTO]):
+        if method not in self.strategies:
+            raise ValueError("Invalid login method")
```
Avariq, line 108: Don't forget to specifically catch this exception later, so that you don't accidentally return status code 500 on API

service/api_logic/user_logic.py
```diff
@@ -92,34 +103,16 @@ def confirm_token(self, token: str, expiration=3600):
...
-
-        if not bcrypt.checkpw(password.encode('utf-8'), user.password_hash.encode('utf-8')):
-            self._logger.warning("Passwords do not match")
-            raise IncorrectPasswordError()
-
-        token = await self.__generate_auth_token(user)
-
-        return OutputLogin(email = user.email, token = token, id = user.user_id)
+    async def log_in(self, method: str, credentials: Union[InputUserLoginDTO, InputUserByGoogleDTO]):
+        if method not in self.strategies:
```
Avariq, line 107: BTW:

I'd move this strategy-selection logic somewhere closer to the strategies themselves.
It's not good to check wether the strategy is correct for log_iin and sign_up separately if we can do that in one place. Maybe let AuthContext decide which strategy has been selected. It'd become more of a factory then, but general idea remains

dto/api_input.py
```diff
@@ -113,9 +119,9 @@ class NewPasswordDTO(BaseDTO):
     email = fields.Str(required=True)
 
 
-class InputUserLoginDTO(BaseDTO):
-    email_or_username = fields.Str(required=True)
-    password_hash = fields.Str(required=True)
+# class InputUserLoginDTO(BaseDTO):
+#     email_or_username = fields.Str(required=True)
+#     password_hash = fields.Str(required=True)
```
Iarynovskyi, line 125: remove comment

exept/exeptions.py
```diff
@@ -69,7 +69,7 @@ def __init__(self):
         super().__init__(message)
 
 
-class IncorrectPasswordError(CustomQSportException):
+class IncorrectUserDataError(CustomQSportException):
```
Avariq, line 72: Don't we need some message here as well? (along with status_code)

exept/exeptions.py
```diff
@@ -83,4 +83,19 @@ class UserAlreadyExistError(CustomQSportException):
...
+
+class IncorrectLogInStrategyMethod(CustomQSportException):
+    status_code = 404
+
+    def __init__(self, data_value):
+        message = f"{data_value} is not a log in method in this app"
+        super().__init__(message)
+
+
+class InvalidDataForGoogleLogIn(CustomQSportException):
```
Avariq, line 96: Imagine we have Facebook, GitHub, Twitter (X) and meny other Auth providers. If we were to create an exception for each one of these how would they differ? Same status code, message differs only by auth_provider's name. So it's probably better to create an exception that is a bit more generic, like:

`InvalidAuthenticationDataError`

which would have: `message = f'{auth_provider_name} data is invalid'`

exept/exeptions.py
```diff
@@ -83,4 +83,19 @@ class UserAlreadyExistError(CustomQSportException):
     def __init__(self, date_value):
         self.date_value = date_value
         message = f"User with such data already exist'"
+        super().__init__(message)
+
+class IncorrectLogInStrategyMethod(CustomQSportException):
```
Avariq, line 88: Since it is an exception and not some usual class we have to specify that in name.

```suggestion
class IncorrectLogInStrategyMethodError(CustomQSportException):
```

Do tha for all exceptions you've made to keep it clean

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,96 @@
...
+class AuthManager:
+    def __init__(self, user_service):
+        self._user_service = user_service
+        self.strategies = {
+            AuthMethods.SIMPLE.value: SimpleAuthHandler(user_service=self._user_service),
+            AuthMethods.GOOGLE.value: GoogleAuthHandler(user_service=self._user_service),
+        }
+
+    async def execute_log_in(self, method: str, credentials: T):
+        if method not in self.strategies:
```
Avariq, line 39: `method` -> `strategy`?

Or are they somehow different

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,96 @@
...
+class SimpleAuthHandler(AuthHandler[T]):
+    async def authenticate(self, credentials: T):
+        user = self._user_service.get_user_by_email_or_username(credentials.email_or_username, credentials.email_or_username)
+
+        if not user:
+            self._user_service._logger.warning("User does not exist")
+            raise UserDoesNotExistError(user.email)
+
+        if not bcrypt.checkpw(credentials.password_hash.encode('utf-8'), user.password_hash.encode('utf-8')):
+            self._user_service._logger.warning("Passwords do not match")
```
Avariq, line 55: That's sensitive info, let's not log it. We shouldn't specify anywhere anything about how close/far the user is from logging in. We never know who might get access to our DB or our logs. That info might help somebody hack a user in the future

Avariq: Also make sure we don't return anything like this to user (I mean through API). We don't want to become a meme, like: "This password is already used by user12345" :)

service/api_logic/user_logic.py
```diff
@@ -1,26 +1,33 @@
...
         self._logger = Logger("logger", "all.log").logger
 
 
+    def get_user_by_email_or_username(self, email_front=None, username_front=None):
+        return self._user_dal.get_user_by_email_or_username(email_front, username_front)
+
+
     async def create_user(self, email_front, username_front, password_front):
-        existing_user = self._user_dal.get_user_by_email_or_username(email_front, username_front)
+        existing_user = self.get_user_by_email_or_username(email_front, username_front)
```
Avariq, line 30: MINOR:

Why don't we write a separate query to DB which will just check whether a user exists? That way we can spare some memory.

service/api_logic/user_logic.py
```diff
@@ -67,59 +74,49 @@ def __send_reset_email(self, user: User, token: str):
...
+        user = self.get_user_by_email_or_username(email)
         if not user:
             raise UserDoesNotExistError(email)
 
         return self._serializer.dumps(user.username, salt = "email-confirm")
 
 
-    def reset_user_password(self, email, new_password: str) -> str:
-        user = self._user_dal.get_user_by_email_or_username(email)
+    def reset_user_password(self, email, new_password: str):
```
Avariq, line 84: Don't fix it. Maybe we'll do it in scope of another PR.

Make sure you convey that to others:

There is an order in which we place our methods within our classes. Usually it's private -> protected -> public, meaning private are on top of the class and everything else is below. 

Make sure you know your code conventions for the language you use, since they usually differ from one to another.

service/api_logic/user_logic.py
```diff
@@ -67,59 +74,49 @@ def __send_reset_email(self, user: User, token: str):
...
             raise UserDoesNotExistError(email)
 
         self._user_dal.update_user_password(user, new_password)
-        new_jwt = create_access_token(identity = user.email)
+        new_jwt = create_access_token(identity=user.email)
+        new_refresh = create_refresh_token(identity=user.email)
 
-        return new_jwt
+        response = jsonify({"message": "Password reset successful"})
+        set_access_cookies(response, new_jwt)
```
Avariq, line 94: This is API layer's job

service/api_logic/user_logic.py
```diff
@@ -67,59 +74,49 @@ def __send_reset_email(self, user: User, token: str):
...
+        set_refresh_cookies(response, new_refresh)
+
+        return response
 
 
     def confirm_token(self, token: str, expiration=3600):
         username = self._serializer.loads(token, salt = "email-confirm", max_age = expiration)
-        user = self._user_dal.get_user_by_email_or_username(None, username)
+        user = self.get_user_by_email_or_username(None, username)
```
Avariq, line 102: Why None?

Can't we just do `username=username` or smth?

service/api_logic/user_logic.py
```diff
@@ -67,59 +74,49 @@ def __send_reset_email(self, user: User, token: str):
...
-        if not bcrypt.checkpw(password.encode('utf-8'), user.password_hash.encode('utf-8')):
-            self._logger.warning("Passwords do not match")
-            raise IncorrectPasswordError()
-
-        token = await self.__generate_auth_token(user)
-
-        return OutputLogin(email = user.email, token = token, id = user.user_id)
+    async def log_in(self, credentials: InputUserLogInDTO):
+        login_context = AuthManager(self)
+        user = await login_context.execute_log_in(credentials.auth_provider, credentials)
```
Avariq, line 109: MINOR:

`credentials.auth_provider, credentials` doesn't sound nice does it?
We either just pass "credentials" (whole object), or do some proper inheritance, like:

`class InputUserDto:
    auth_provider: AuthMethods (your auth methods Enum)
    user_credentials: UserCredentials (your dto for creds)
`

then it'll be like: user = await login_context.execute_log_in(nameofdto.auth_provider, nameofdto.user_credentials)

service/api_logic/auth_strategy.py
```diff
@@ -0,0 +1,96 @@
...
+from database.models import User
+from dto.api_input import InputUserLogInDTO
+from dto.api_output import OutputLogin
+from exept.exeptions import IncorrectUserDataError, UserDoesNotExistError, IncorrectLogInStrategyMethod, \
+    InvalidDataForGoogleLogIn
+from typing import Generic, TypeVar
+
+T = TypeVar("T")
+
+class AuthMethods(Enum):
```
Avariq, line 16: Seems to me like something that can be used in many projects of our program (API layer and BusinessLayer at least)

I suggest moving this to some separate place. You can create a separate "project" which would hold all such "global" models

service/api_logic/user_logic.py
```diff
@@ -67,59 +74,49 @@ def __send_reset_email(self, user: User, token: str):
...
-        if not user:
-            user = User(email = email, username = email.split('@')[0])
-            self._user_dal.create_user(user)
+            return self._serializer.dumps(user.email, salt = "user-auth-token")
 
-        token = self.__generate_auth_token(user)
 
-        return OutputLogin(email = user.email, token = token, id = user.user_id)
+    def get_generate_auth_token(self, user):
+        return self.__generate_auth_token(user)
```
Avariq, line 119: __generate_auth_token is async, but you don't await it. Is that normal? (most likely not)

and also I don't think we need two methods that do the same thing. 

That'd be logical if you had:

`async def __generate_auth_token(self, user, salt):`
`    return self._serializer.dumps(user.email, salt = salt)`

and 

`async def get_generate_auth_token(self, user):`
`    return await self.__generate_auth_token(user, salt = "user-auth-token")`

That'd have some sense to it and would allow reusing the private method elsewhere, while keeping the public one intact

Avariq (approved):

PR #20, opened 2025-02-04
UJP-19 news recommendations for user

Iarynovskyi:
for ideal it is need more tests and some work with coefficients


api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+@inject
+@handle_exceptions
+@logger.log_function_call()
+async def get_recommendations_endpoint(
+        service_recs: RecommendationService = Provide[Container.recommendation_service],
+        service_news: NewsService = Provide[Container.news_service],
+    ):
+    try:
+        user_email = get_jwt_identity()
+        rec = await service_recs.get_recommendations_from_db(user_email)
```
Avariq, line 98:
```suggestion
        rec = await service_recs.get_user_recommendations(user_email)
```

no need to specify the data source unless there is a good reason to do so.

just KISS, YAGNI and so on (Google search time)

api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+@jwt_required()
+@inject
+@handle_exceptions
+@logger.log_function_call()
+async def get_recommendations_endpoint(
+        service_recs: RecommendationService = Provide[Container.recommendation_service],
+        service_news: NewsService = Provide[Container.news_service],
+    ):
+    try:
+        user_email = get_jwt_identity()
```
Avariq, line 97: What if somebody tries to get here with invalid jwt token? Are we guaranteed that `user_email` will be filled with data?

api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+        rec = await service_recs.get_recommendations_from_db(user_email)
+        all_recs_with_data = await service_news.send_all_info_from_blob_to_user(rec)
```
Avariq, line 99: There is no "info" in our project. We work with a predefined set of entities.
Also "send" is not the best word here, since we're getting data. It might be confusing to the reader.

Avariq: Also, can't our news_service do all that stuff?
Do we really need to call two services here?

I don't think we actually need a recommendations service. DAL should suffice. That way, let's say, our news_service will have its logic for "get_user_recommendations" and streams_service will have another implementation of that or similar method (if we ever decide to do recs for streams as well that is). 

It'd be better to just have our service do all the job. Just plain `news_service.get_user_recs` should suffice.

Avariq: Maybe we should have our UserService return the preferences and NewsService return news based on those preferences

Avariq: **NOTE:**
If there is a need of using multiple services at once but you don't want to create any dependecies between them (like: I don't think NewsService should depend on UserService or vice versa, right?) - consider creating another layer of abstraction: some new class, sort of manager, that will stand in between API and our Services.

I am not saying we need this here, I'm just saying you can think that way too

api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+
+@news_app.route("/fast/recommendation", methods=["POST"])
+@inject
+@handle_exceptions
+@logger.log_function_call()
+async def fast_generate_recommendation_endpoint(service_recs: RecommendationService = Provide[Container.recommendation_service]):
+    try:
+        data = request.get_json()
+        dto = InputUserByIdDTO().load(data)
+        rec = await service_recs.fast_hybrid_recommendation_system(dto)
```
Avariq, line 116: `fast_hybrid_recommendation_system` is a bad name. I don't understand what it does by reading its name. Same goes for the variable name

api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+        return get_custom_error_response(e)
+
+
+@news_app.route("/all/recommendation", methods=["GET"])
+@inject
+@handle_exceptions
+@logger.log_function_call()
+async def all_generate_recommendation_endpoint(service_recs: RecommendationService = Provide[Container.recommendation_service]):
+    try:
+        rec = await service_recs.auto_hybrid_recommendation_system()
```
Avariq, line 130: same here

api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+    except CustomQSportException as e:
+        logger.error(f"Error in Get Recommendations /: {str(e)}")
+        return get_custom_error_response(e)
+
+
+@news_app.route("/all/recommendation", methods=["GET"])
+@inject
+@handle_exceptions
+@logger.log_function_call()
+async def all_generate_recommendation_endpoint(service_recs: RecommendationService = Provide[Container.recommendation_service]):
```
Avariq, line 128: endpoint name is not the best either

api/routes/api_news.py
```diff
@@ -73,3 +82,54 @@ def specific_article(service: NewsService = Provide[Container.news_service]):
...
+    except CustomQSportException as e:
+        logger.error(f"Error in Get Recommendations /: {str(e)}")
+        get_custom_error_response(e)
+
+
+@news_app.route("/fast/recommendation", methods=["POST"])
+@inject
+@handle_exceptions
+@logger.log_function_call()
+async def fast_generate_recommendation_endpoint(service_recs: RecommendationService = Provide[Container.recommendation_service]):
```
Avariq, line 112: endpoint name is not the best either

ALSO, I don't see a reason why y'all keep adding "endpoint" in the end of the name
Seems a little redundant

database/postgres/dal/news.py
```diff
@@ -27,3 +27,19 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
     def get_news_by_id(self, blob_id: str):
         return self.session.query(News).filter(News.blob_id == blob_id).first()
 
+
+    def get_news_by_real_id(self, news_id: str):
```
Avariq, line 31: real id? I guess it refers to our Database id. Consider renaming to "native_id" or "db_id". Naming is very crucial. Sometimes it takes me more time to come up with an alrorithm than to name all the variables and methods

database/postgres/dal/news.py
```diff
@@ -27,3 +27,19 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
...
+        return self.session.query(News).filter(News.news_id == news_id).first()
+
+
+    def get_sport_by_name(self, sport_name):
```
Avariq, line 35: Does it belong here? Maybe we'd better create SportsDal or smth?

Single responsibility principle. (Unless you meant something else by "sport" here)

database/postgres/dal/news.py
```diff
@@ -27,3 +27,19 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
...
+        return self.session.query(Sport).filter_by(sport_name = sport_name).first()
+
+
+    def get_news_ids(self, time_limit):
```
Avariq, line 39: I see `get_news_ids` but I don't understand what news? What ids? Judging by the argument of the method we expect some time limit (which is also a very vague name).

`get_news_by_time_limit` for your case, but 'time_limit' is a vague name, so specify what exactly does it mean.

database/postgres/dal/news.py
```diff
@@ -27,3 +27,19 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
...
+        return (
+            self.session.query(News.news_id)
+            .filter(News.save_at >= time_limit)
+            .all()
```
Avariq, line 43: What if we have thousands of news? Pagination is required.

However, that's no use to hardcode pagination in each method like this one inside every DAL we have.
We are getting very close to SearchEngine here, so let's discuss it on the meeting today

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+        if not users_that_want_recs:
+            raise NoUsersInDBError
+
+        interaction_df = self.__get_interactions(users_that_want_recs)
+        users_interaction_matrix = self.__create_interaction_matrix(interaction_df)
+        for user in users_that_want_recs:
+            try:
+                recommendations_for_user_with_score = await self.__user_based_recommendations(user.user_id, users_interaction_matrix, top_n=top_n)
+                if not recommendations_for_user_with_score or not any(recommendations_for_user_with_score):
+                    raise EmptyRecommendationListForUserError(user.user_id)
```
Avariq, line 41: We need to talk about it on meeting today.

Concerns:
1) All users query
2) Many requests
3) Bad method name
4) Do we need an exception here?

database/postgres/dal/recommendation.py
```diff
@@ -0,0 +1,110 @@
...
+    UserPreference, Sport, TeamInNews, UserRecommendations
+from sqlalchemy import union_all, literal, func
+
+
+class RecommendationDAL:
+    def __init__(self, session=None):
+        self.session = session
+
+
+    def get_all_users(self):
```
Avariq, line 11: Seems like this one belongs to UserDAL and has nothing to do with Recs.

I'm also concerned with ".all()"

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+    def __init__(self, recommendation_dal, user_dal, news_dal):
+        self._recommendation_dal = recommendation_dal
+        self._user_dal = user_dal
+        self._news_dal = news_dal
+        self.logger = Logger("logger", "all.log").logger
+        self._recommendations_list = []
+
+
+    async def auto_hybrid_recommendation_system(self, top_n=BASE_LIMIT_OF_NEWS_FOR_RECS):
+        users_that_want_recs = self._recommendation_dal.get_all_users()
```
Avariq, line 31: We just return all users here, so `users_that_want_recs` name is misleading

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+        if not users_that_want_recs:
+            raise NoUsersInDBError
```
Avariq, line 33: Discuss: Do we need an exception here?

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+        users_interaction_matrix = self.__create_interaction_matrix(interaction_df)
+        for user in users_that_want_recs:
+            try:
+                recommendations_for_user_with_score = await self.__user_based_recommendations(user.user_id, users_interaction_matrix, top_n=top_n)
+                if not recommendations_for_user_with_score or not any(recommendations_for_user_with_score):
+                    raise EmptyRecommendationListForUserError(user.user_id)
+
+                self._recommendations_list.append(recommendations_for_user_with_score)
+            except EmptyRecommendationListForUserError as e:
+                self.logger.warning(f"{e}, so skipping this user")
```
Avariq, line 45:
```suggestion
                self.logger.debug(f"{e}, so skipping this user")
```

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+                continue
+
+        self.__save_to_db_users_recommendations(self._recommendations_list)
```
Avariq, line 48: Can we save those recs in some other place except our DB? If not then just `save_user_rec`

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+
+        return self._recommendations_list
+
+
+    def __get_interactions(self, users_that_want_recs, time_limit=MAX_LIMIT_FOR_REC_IN_DAYS):
+        time_limit = datetime.now() - timedelta(days=time_limit)
+        interactions = self._recommendation_dal.get_user_interactions(time_limit)
+        interactions_df = pd.DataFrame(interactions, columns=['user_id', 'news_id', 'interaction', 'timestamp'])
```
Avariq, line 70: Can't we do that on DB level?

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+        interactions_grouped = interactions_df.groupby(['user_id', 'news_id'], as_index=False)['interaction'].sum()
```
Avariq, line 71: DB level?

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+
+        all_news_by_last_const_days = self._news_dal.get_news_ids(time_limit)
+        all_news_by_last_const_days_df = pd.DataFrame(all_news_by_last_const_days, columns=['news_id'])
+
+        if isinstance(users_that_want_recs, list):
```
Avariq, line 76: It's our backend and we control everything that's happening here. Can't we just ensure, that this method will always get a list?

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+            user_ids = [user.user_id for user in users_that_want_recs]
+        else:
+            user_ids = [users_that_want_recs.user_id]
+
+        all_users_df = pd.DataFrame(user_ids, columns=['user_id'])
+
+        all_combinations = all_users_df.merge(all_news_by_last_const_days_df, how='cross')
```
Avariq, line 83: Potential OutOfMemoryException incoming! Brace for the impact!

service/api_logic/recommendation_logic.py
Iarynovskyi: I have not tested it in practice, but it seems like that logic may break somewhere later

service/api_logic/recommendation_logic.py
```diff
@@ -0,0 +1,289 @@
...
+        return interactions_df.pivot_table(
+            index='user_id',
+            columns='news_id',
+            values='interaction',
+            aggfunc='sum',
+            fill_value=0
+        )
+
+
+    async def __user_based_recommendations(self, user_id, user_interaction_matrix, top_n=BASE_LIMIT_OF_NEWS_FOR_RECS):
```
Avariq, line 102: Discuss this one

Potential overkill

service/api_logic/recommendation_logic.py
Iarynovskyi: I did not know where to work with the dataframes. if at the database level so then I will move it there

database/postgres/dal/news.py
Avariq: Pagination?

database/postgres/dal/news.py
```diff
@@ -27,3 +30,117 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
     def get_news_by_id(self, blob_id: str):
         return self.session.query(News).filter(News.blob_id == blob_id).first()
 
+
+    def get_news_by_native_id(self, news_ids):
+        return self.session.query(News).filter(News.news_id.in_(news_ids)).all()
```
Avariq, line 35: Pagination?

UPD:
We'll handle it when the search Engine is done

database/postgres/dal/news.py
```diff
@@ -27,3 +30,117 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
...
+            all_info_about_news_with_user_interactions_with_them
+            .filter(
+                News.save_at >= period_of_time
+            )
+        )
+
+        return list(set(all_info_about_news_by_period_of_time_with_user_interactions_with_them.all()))
+
+
+    def data_frame_to_work_with_user_and_news_data(self, user_and_news: list[tuple]) -> pd.DataFrame.mask and pd.DataFrame.mask and pd.DataFrame:
```
Avariq, line 92: Should it be public?

database/postgres/dal/news.py
```diff
@@ -27,3 +30,117 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
...
+            'sport_score': 'first',
+            'team_score': 'sum',
+        }).reset_index()
+
+        news_coefficients_without_duplicates = news_coefficients_without_duplicates.set_index('news_id')
+
+        return news_coefficients_without_duplicates
+
+
+    def assign_adjusted_scores_for_masks(
```
Avariq, line 124: probably a private method again

database/postgres/dal/news.py
Avariq: And also, are we guaranteed the amount of memory used for creating this dataframe would be somewhat manageble (worst case scenario)?

database/postgres/dal/news.py
```diff
@@ -27,3 +30,117 @@ def fetch_news(self, order_by=None, limit=None, filters=None):
...
+    ) -> pd.DataFrame.mask and pd.DataFrame.mask:
+
+        user_interact_with_this_news_mask[['adjusted_score','blob_id']] = news_coefficients_without_duplicates[['adjusted_score','blob_id']]
+        user_not_interact_with_this_news_mask[['adjusted_score','blob_id']] = news_coefficients_without_duplicates[['adjusted_score','blob_id']]
+        self.delete_dataframe(news_coefficients_without_duplicates)
+
+        return user_interact_with_this_news_mask, user_not_interact_with_this_news_mask
+
+
+    def delete_dataframe(self, dataframe):
```
Avariq, line 138: First - a private method for sure
Second - isn't there a better option? Like using "with" or smth

database/postgres/dal/sport.py
```diff
@@ -4,7 +4,7 @@
 from typing import Optional
 
 class SportDAL:
-    def __init__(self, db_session: Session):
+    def __init__(self, db_session = None):
```
Avariq, line 7: ?

database/postgres/dal/user.py
```diff
@@ -28,4 +29,24 @@ def get_user_by_id(self, user_id: int) -> User:
 
     def update_user_password(self, user: User, new_password):
         user.password_hash = new_password
-        self.session.commit()
\ No newline at end of file
+        self.session.commit()
+
+    def get_user_id_be_email(self, email: str) -> User:
```
Avariq, line 34: get user ID but return the whole user.

Either rename or return just an ID (select only ID from database, don't do that in python)

database/postgres/dal/user.py
```diff
@@ -28,4 +29,24 @@ def get_user_by_id(self, user_id: int) -> User:
...
+        return self.session.query(User).filter(User.email == email).first()
+
+    def get_all_users(self):
+        return self.session.query(User).all()
```
Avariq, line 38: That's A LOT of memory here. I doubt we truly need this

database/session.py
```diff
@@ -7,5 +7,13 @@
...
+engine = create_engine(
+    DATABASE_URL,
+    echo=True,
+    pool_size=20,
+    max_overflow=10,
+    pool_timeout=30,
+    pool_recycle=180
+)
+
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
```
Avariq, line 19: I'm curios about how that works

database/postgres/dal/news.py
Iarynovskyi: With 20 news it's used newar 6 MB, and our parser doesn’t save a lot of news, soo.. i don't know amount of memory used in worst case scenario

dto/api_input.py
```diff
@@ -127,6 +127,10 @@ class GetUserPreferencesDTO(BaseDTO):
     user_id = fields.Int(required=False, missing=None)
 
 
+class InputUserByIdDTO(BaseDTO):
```
Avariq, line 130: We can always get ID from JWT, but that's fine as well

dto/api_input.py
```diff
@@ -127,6 +127,10 @@ class GetUserPreferencesDTO(BaseDTO):
...
+    user_id = fields.Int(required=False, missing=None)
```
Avariq, line 131: Can our id be less than 0?

database/postgres/dal/sport.py
Iarynovskyi: This is an old method that was used a little differently before, so I changed it

database/postgres/dal/user.py
Iarynovskyi: I must removed this, it is not used. I forgot delete..

exept/exeptions.py
```diff
@@ -83,4 +83,28 @@ class UserAlreadyExistError(CustomQSportException):
     def __init__(self, date_value):
         self.date_value = date_value
         message = f"User with such data already exist'"
+        super().__init__(message)
+
+
+class NoUsersInDBError(CustomQSportException):
```
Avariq, line 89: I would do just "UserNotFound"

This exception is basically impossible to get anyways. On real project you'd have AT LEAST some developer accounts in place or smth like that. Plust it'd expose we have no users to our API clients  :)

dto/api_input.py
Iarynovskyi: there is a negativity check in the base class

exept/exeptions.py
```diff
@@ -83,4 +83,28 @@ class UserAlreadyExistError(CustomQSportException):
...
+    status_code = 404
+
+    def __init__(self):
+        message = f"There are no users in database"
+        super().__init__(message)
+
+
+class EmptyRecommendationListForUserError(CustomQSportException):
```
Avariq, line 97: Line 89: NoUser....


Line 97: EmptyRecc...

Keep following the same naming pattern EVERYWHERE. "NoSomethingFound" is preferred of course

Avariq: "NoRecommendationsFound" sounds better

exept/exeptions.py
```diff
@@ -83,4 +83,28 @@ class UserAlreadyExistError(CustomQSportException):
...
+    status_code = 204
+
+    def __init__(self, user_id):
+        message = f"There are no recommendations for user '{user_id} at this moment'"
+        super().__init__(message)
+
+
+class NoJWTInCookieError(CustomQSportException):
```
Avariq, line 105: That's basically "UserNotLoggedIn" or smth

We don't care what happened to the JWT - if it's corrupted/missing/fake - User is not logged in and may as well go fuck themself

service/api_logic/helpers/calculating_helper.py
```diff
@@ -0,0 +1,80 @@
...
+    MAX_LIMIT_FOR_REC_IN_DAYS = 21
+    SCORE_FOR_NOT_PREFER_SPORT = 0.25
+    SCORE_FOR_PREFER_SPORT = 0.20
+    SCORE_FOR_NOTHING = 0.1
+    SCORE_FOR_PREFER_TEAM = 0.2
+    BASE_LIMIT_OF_NEWS_FOR_RECS = 5
+    MIN_PROMOTION_COF = 1
+    MAX_PROMOTION_COF = 4
+
+class CalculatingHelper:
```
Avariq, line 19: I bet we calculate something very specific, so the name should align with that

service/api_logic/helpers/calculating_helper.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+    def __calculate_sport_score_vectorized(self, news_sports: list[int], preferred_sports: list[int], user_interactions: pd.DataFrame) -> pd.DataFrame:
+        preferred_count = np.sum(np.isin(news_sports, preferred_sports)) * RecommendationConsts.SCORE_FOR_PREFER_SPORT.value
+        interacted_count = np.sum(np.isin(news_sports, user_interactions)) * RecommendationConsts.SCORE_FOR_NOT_PREFER_SPORT.value
+
+        return preferred_count + interacted_count or RecommendationConsts.SCORE_FOR_NOTHING.value
+
+
+    def calculating_news_coefficients_by_parameters_from_df_and_user_preferences(
```
Avariq, line 38:
```suggestion
    def calculate_news_coefficients_by_parameters_from_df_and_user_preferences(
```

Avariq: A bit too long of a name, but let it be

service/api_logic/helpers/calculating_helper.py
```diff
@@ -0,0 +1,80 @@
...
+                user_interact_with_this_news_mask['sport_id']
+            )
+        )
+
+        user_and_news_details_df.drop(columns=['sport_id', 'save_at', 'team_id'], inplace=True)
+
+        return user_and_news_details_df
+
+
+    def calculating_adjusted_score_by_coefficients_inside_df(self, dataframe: pd.DataFrame) -> pd.DataFrame:
```
Avariq, line 71:
```suggestion
    def calculate_adjusted_score_by_coefficients_inside_df(self, dataframe: pd.DataFrame) -> pd.DataFrame:
```

service/api_logic/helpers/calculating_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        dataframe['adjusted_score'] = dataframe['interest_rate_score'] * dataframe['time_score'] * dataframe['team_score'] * dataframe['sport_score']
+        return dataframe
+
+
+    def calculate_user_average_adjusted_score_by_last_watch_news(self, interacted_df: pd.DataFrame, not_interacted_df: pd.DataFrame) -> pd.DataFrame:
```
Avariq, line 76: And here you actually went with `calculatE`... Whatever you drink or smoke - you'd better stop, mate

service/api_logic/helpers/calculating_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        return (
+            interacted_df['adjusted_score'].sum() / len(not_interacted_df)
```
Avariq, line 78: don't understand a thing here, but make sure we don't get division by zero here

service/api_logic/managers/recommendation_menager.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return self.__final_user_recommendations(
+             user_not_saw_this_news_df,
+             recs_by_user_preferences,
+             recs_by_user_last_watch,
+             user_id
+        )
+
+
+    def __final_user_recommendations(
```
Avariq, line 32: get? delete?
please be specific

service/api_logic/managers/recommendation_menager.py
```diff
@@ -0,0 +1,70 @@
...
+            recs_by_user_preferences: list[int],
+            recs_by_user_last_watch: pd.DataFrame,
+            user_id: int
+    ):
+        recommendations_list_by_user_preferences = [
+            {
+                "news_id": news,
+                "score": user_not_saw_this_news_df.loc[news, 'adjusted_score'],
+                "user_id": user_id,
+                "article": self.__retrieve_user_recommendations_from_blob(user_not_saw_this_news_df.loc[news, 'blob_id'])
```
Avariq, line 44: Many different blob requests? Can we do that in one request instead? Or at least some batches, because blob is quite slow

service/api_logic/managers/recommendation_menager.py
```diff
@@ -0,0 +1,70 @@
...
+            }
+            for news in recs_by_user_preferences
+        ]
+        recommendations_list_by_user_last_watch = [
+            {
+                "news_id": news,
+                "score": user_not_saw_this_news_df.loc[news, 'adjusted_score'],
+                "user_id": user_id,
+                "article":  self.__retrieve_user_recommendations_from_blob(user_not_saw_this_news_df.loc[news, 'blob_id'])
```
Avariq, line 53: this method could be very slow

service/api_logic/managers/recommendation_menager.py
```diff
@@ -0,0 +1,70 @@
...
+class RecommendationManager:
+    def __init__(self, user_service: UserService, news_service: NewsService):
+        self.user_service = user_service
+        self.news_service = news_service
+        self.calculating_helper = CalculatingHelper()
+
+    def get_recommended_news_for_user(self, user_id: int):
+        preferred_teams, preferred_sports = self.user_service.get_user_sport_and_club_preferences(user_id)
+
+        user_not_saw_this_news_df, recs_by_user_preferences, recs_by_user_last_watch = self.news_service.user_recommendations_based_on_preferences_and_last_watch(
```
Avariq, line 17:
```suggestion
        user_unseen_news_df, recs_by_user_preferences, recs_by_user_last_watch = self.news_service.user_recommendations_based_on_preferences_and_last_watch(
```

service/api_logic/news_logic.py
```diff
@@ -1,29 +1,36 @@
...
 from database.azure_blob_storage.save_get_blob import blob_get_news
-from sqlalchemy.sql.expression import ClauseElement
-from exept.handle_exeptions import handle_exceptions
-from service.api_logic.scripts import get_sport_by_name
-from database.session import SessionLocal
 from logger.logger import Logger
+from service.api_logic.helpers.calculating_helper import CalculatingHelper
+
+
+class LimitsConsts(Enum):
```
Avariq, line 12: For the future: This should be in some separate place. All your general stuff should be separated

Don't fix it, let it be

service/implementation/auto_request_api/shedulers.py
```diff
@@ -2,40 +2,73 @@
 from apscheduler.schedulers.background import BackgroundScheduler
 from concurrent.futures import ThreadPoolExecutor
 from service.implementation.auto_request_api.logic_auto_request import apis, auto_request_system, keep_db_alive
+from logger.logger import Logger
 
-scheduler1 = BackgroundScheduler()
-scheduler2 = BackgroundScheduler()
 
-executor = ThreadPoolExecutor(max_workers=3)
+class AutoSchedulerService:
```
Avariq, line 8: That's probably not a part of this PR..

PR #21, opened 2025-02-10
UJP-143 Feature/recommendation email sender system

IceStorman:


database/alembic.ini
```diff
@@ -61,7 +61,7 @@ version_path_separator = os  # Use os.pathsep. Default configuration used for ne
 # are written from script.py.mako
 # output_encoding = utf-8
 
-sqlalchemy.url = postgresql://mac:D100406m@localhost:5432/SoftProg
+sqlalchemy.url = postgresql://postgres:mysql0306@localhost:5432/SoftServeProject
```
AndrewKozovyi: Don't change this!

Taras4568, line 64: no address in commit

database/azure_blob_storage/save_get_blob.py
```diff
@@ -299,6 +300,8 @@ def save_news_index_to_db(blob_name: str, json_data,  session) -> None:
                             team_index_id=team_index_id
                         )
                         session.add(team_index)
+
+                        subscription_manager.try_add_subscribers_to_temp_table(team_index.team_index_id)
```
AndrewKozovyi, line 304: naming, why it tries and do nothig, if it fails

database/migrations/versions/6b62e4083dc9_added_tempsubscribersdata_table.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision: str = '6b62e4083dc9'
+down_revision: Union[str, None] = '42e239ddb722'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
```
Iarynovskyi, line 21: two head now..... womp womp

database/migrations/versions/6b62e4083dc9_added_tempsubscribersdata_table.py
```diff
@@ -0,0 +1,36 @@
...
+    sa.ForeignKeyConstraint(['team_ids'], ['TeamIndex.team_index_id'], ),
+    sa.PrimaryKeyConstraint('id')
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('TempSubscribersData')
+    # ### end Alembic commands ###
```
AndrewKozovyi, line 36: 2 heads for db...

database/migrations/versions/6b62e4083dc9_added_tempsubscribersdata_table.py
Taras4568, line 1: migrations on db branch 'cause we now have 2 heads

database/models/temp_subscribers_data.py
Taras4568, line 1: on fucking database branch

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,43 @@
+from dask.array import delete
+from sqlalchemy.orm import Session
+
+from database.models import User, TempSubscribersData, UserClubPreferences
+
+
+class UserSubscriptionDAL:
+    def __init__(self, db_session: Session):
```
Iarynovskyi, line 8: better create this DAL in DI container

service/implementation/email_sender/user_subscription_manager.py
AndrewKozovyi, line 1: why not using our container?

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,43 @@
...
+        new_rows = [
+            TempSubscribersData(team_ids=preference_index, subscriber_emails=user.email)
+            for user in users
+        ]
+
+        self.db_session.add_all(new_rows)
+        self.db_session.commit()
+
+    def __get_users_by_preference_index(self, preference_index):
+        query = (
```
Iarynovskyi, line 23: This should be in UserDal or PreferencesDal

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,43 @@
...
+
+        new_rows = [
+            TempSubscribersData(team_ids=preference_index, subscriber_emails=user.email)
+            for user in users
+        ]
+
+        self.db_session.add_all(new_rows)
+        self.db_session.commit()
+
+    def __get_users_by_preference_index(self, preference_index):
```
Taras4568, line 22: move to fucking preference dal

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        creds = None
+
+        if os.path.exists("token.json"):
+            creds = Credentials.from_authorized_user_file("token.json", self.__SCOPES)
+
+        if not creds or not creds.valid:
+            if creds and creds.expired and creds.refresh_token:
+                creds.refresh(Request())
+            else:
+                credentials_path = r"C:\Users\LEGION\Documents\Softserve\SoftServeProject\credentials.json"
```
AndrewKozovyi, line 88: do you really think it's okay?

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+from database.session import SessionLocal
+
+from pathlib import Path
+
+from dotenv import load_dotenv
+
+from sqlalchemy import event
+
+class UserSubscriptionManagerMeta(type):
+    _instances = {}
```
Iarynovskyi, line 22: I think that better to create Singleton service in DI

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+            _instance = super().__call__(*args, **kwargs)
+            cls._instances[cls] = _instance
+        return cls._instances[cls]
+
+class UserSubscriptionManager(metaclass=UserSubscriptionManagerMeta):
+    __SCOPES = ["https://www.googleapis.com/auth/gmail.send"]
+
+    def __init__(self):
+        self.__session = SessionLocal()
+
```
Iarynovskyi, line 35: BL should not have access to the DL
So move it from here

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        current_dir = Path(__file__).resolve().parent if "__file__" in locals() else Path.cwd()
+        envars = current_dir / ".env"
+        load_dotenv(envars)
+
```
Iarynovskyi, line 39: you can do better

` app.config['MAIL_USERNAME'] = os.getenv('MAIL_USERNAME')
    
app.config['MAIL_PASSWORD'] = os.getenv('MAIL_PASSWORD')`

maybe something like this. And moved this to app.config
And we have already gmail for mailings

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        self.__sender_email = os.getenv("EMAIL")
+        self.__sender_password = os.getenv("PASSWORD")
+
+        event.listen(TempSubscribersData, "after_insert", self.__on_subscribers_inserted)
+
+    def try_add_subscribers_to_temp_table(self, team_index):
+        self.__user_subscription_dal.try_add_subscribers_data(team_index)
+
+    def __on_subscribers_inserted(self, mapper, connection, target):
+        self.__try_send_email_to_users()
```
Iarynovskyi, line 51: maybe it's better to use one function instead of a function that starts something from BL ?

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __try_send_email_to_users(self):
+        subscribed_users = self.__user_subscription_dal.get_subscribed_users_data_and_delete_rows()
+
+        email_subject = "News notification"
+        email_body = "The news with the team that you subscribed in have just been released"
+
```
Iarynovskyi, line 58: hardcode

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        message["subject"] = subject
+        raw_message = base64.urlsafe_b64encode(message.as_bytes()).decode()
+
+        send_message = service.users().messages().send(
+            userId="me",
+            body={"raw": raw_message}
+        ).execute()
+
+        print(f"Email sent successfully! Message ID: {send_message['id']}")
+
```
Iarynovskyi, line 77: delete print

Iarynovskyi (changes requested):
Do changes

AndrewKozovyi (changes requested):

api/container/container.py
```diff
@@ -50,5 +52,4 @@ class Container(containers.DeclarativeContainer):
 
 
 
-
-
+    user_subscription_dal = providers.Factory(UserSubscriptionDAL, session=db_session)
```
Avariq, line 55: too many blank lines above

database/azure_blob_storage/save_get_blob.py
```diff
@@ -9,6 +9,7 @@
 from exept.colors_text import print_error_message, print_good_message
 import re
 
+#from service.implementation.email_sender.user_subscription_manager import UserSubscriptionManager
```
Avariq, line 12: ?

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
+from database.models import User, TempSubscribersData, UserClubPreferences
+
+class UserSubscriptionDAL:
+    def __init__(self, db_session):
+        self.db_session = db_session
+
+    def try_add_subscribers_data(self, preference_index):
```
Avariq, line 7: `try_...` naming suggests some response which would allow one to determine whether the "try" was successful or not.

If you want to stick to the naming - make sure the method returns something always (like None, when it failed)

Or, even better, remove the "try_" poart from naming and make sure BL catches exception if any and rethrows something we can understand (like custome exception) or just None or whatever. Because currently we have unhandled exception here if `commit` fails

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+        users = self.__get_users_by_preference_index(preference_index)
+
+        new_rows = [
+            TempSubscribersData(team_ids=preference_index, subscriber_emails=user.email)
```
Avariq, line 11: `team_ids = preference_index`. Looks a little bit odd. Could you please elaborate?

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+        query = (
+            self.db_session.query(User)
+            .join(UserClubPreferences, User.user_id == UserClubPreferences.users_id)
+            .filter(UserClubPreferences.preferences == preference_index)
+            .all()
+        )
+
+        return query
+
+    def get_subscribed_users_data_and_delete_rows(self):
```
Avariq, line 28: rename this method as suggested above.

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+        rows_to_delete = self.db_session.query(TempSubscribersData).all()
```
Avariq, line 29: The main purpose is to get subscribed users to send the emails, right? But the naming suggest we are focused on deleting records. Misleading

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+
+        subscribed_users = list()
+
+        for row in rows_to_delete:
```
Avariq, line 33: Why not use list generator?

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+            copied_user = TempSubscribersData(team_ids=row.team_ids, subscriber_emails=row.subscriber_emails)
+            subscribed_users.append(copied_user)
+
+        self.db_session.query(TempSubscribersData).delete(synchronize_session=False)
```
Avariq, line 37: `synchronize_session = False`

how would it impact our app?

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+from pathlib import Path
+from dotenv import load_dotenv
+
+from sqlalchemy import event
+
+from api.container.container import Container
+from database.postgres.dal.user_subscription import UserSubscriptionDAL
+
+
+class UserSubscriptionManagerMeta(type):
```
Avariq, line 22: I already saw pretty much the same chunk of code before. Do we need that duplication?

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            _instance = super().__call__(*args, **kwargs)
+            cls._instances[cls] = _instance
+        return cls._instances[cls]
+
+class UserSubscriptionManager(metaclass=UserSubscriptionManagerMeta):
+    __SCOPES = ["https://www.googleapis.com/auth/gmail.send"]
```
Avariq, line 32: I don't like hardcoded strings. Hard to maintain. Let's discuss this tomorrow

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+        current_dir = Path(__file__).resolve().parent if "__file__" in locals() else Path.cwd()
+        envars = current_dir / ".env"
+        load_dotenv(envars)
+
+        self.__sender_email = os.getenv("EMAIL")
+        self.__sender_password = os.getenv("PASSWORD")
+
+        event.listen(TempSubscribersData, "after_insert", self.__on_subscribers_inserted)
+
+    @inject
```
Avariq, line 47: What is that `inject` for?

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    def try_add_subscribers_to_temp_table(self, team_index):
+        self.__user_subscription_dal.try_add_subscribers_data(team_index)
+
+    def __on_subscribers_inserted(self, mapper, connection, target):
+        self.__try_send_email_to_users()
+
+    def __try_send_email_to_users(self):
+        subscribed_users = self.__user_subscription_dal.get_subscribed_users_data_and_delete_rows()
```
Avariq, line 55: If you have to write word "AND" in the name of your function: that's like 99.9% chance your method is not SOLID.

Single responsibility, pls

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+            userId="me",
+            body={"raw": raw_message}
+        ).execute()
+
+        print(f"Email sent successfully! Message ID: {send_message['id']}")
+
+    def __authenticate_gmail(self):
+        creds = None
+
+        if os.path.exists("token.json"):
```
Avariq, line 82: What does this do?

service/implementation/email_sender/user_subscription_manager.py
Avariq: I agree with Andrew. Not only it's a hardcoded string but also one that won't work for anybody except you

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+            creds = Credentials.from_authorized_user_file("token.json", self.__SCOPES)
+
+        if not creds or not creds.valid:
+            if creds and creds.expired and creds.refresh_token:
+                creds.refresh(Request())
+            else:
+                credentials_path = r"C:\Users\LEGION\Documents\Softserve\SoftServeProject\credentials.json"
+                flow = InstalledAppFlow.from_client_secrets_file(credentials_path, self.__SCOPES)
```
Avariq, line 90: We absolutely MUST not work with Google Auth on our server. 


Here's some info from our GPT friend:

No, InstalledAppFlow is for desktop or local apps, not Web APIs.

For a Web API, you should use Authorization Code Flow with google-auth and oauthlib. You'd:

Redirect users to Google's OAuth consent page.
Handle the callback with an authorization code.
Exchange the code for access tokens on your backend.
Use google-auth-oauthlib.flow.Flow.from_client_secrets_file(...) instead, with Flow.authorization_url() and Flow.fetch_token().

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,99 @@
...
+                creds = flow.run_local_server(port=0)
+
+            with open("token.json", "w") as token:
+                token.write(creds.to_json())
+
+        return creds
+
+manager = UserSubscriptionManager()
+manager.try_add_subscribers_to_temp_table(1)
```
Avariq, line 99: ?

database/alembic.ini
```diff
@@ -61,7 +61,7 @@ version_path_separator = os  # Use os.pathsep. Default configuration used for ne
 # are written from script.py.mako
 # output_encoding = utf-8
 
-sqlalchemy.url = postgresql://mac:D100406m@localhost:5432/SoftProg
+sqlalchemy.url = postgresql://default-username:password@localhost:5432/project-name
```
AndrewKozovyi, line 64: default-username -> postgres

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+            self.db_session.query(User)
+            .join(UserClubPreferences, User.user_id == UserClubPreferences.users_id)
+            .filter(UserClubPreferences.preferences == preference_index)
+            .all()
+        )
+
+        return query
+
+    def get_subscribed_users_data_and_delete_rows(self):
+        rows_to_delete = self.db_session.query(TempSubscribersData).all()
```
AndrewKozovyi, line 29: dal level

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+
+        subscribed_users = list()
+
+        for row in rows_to_delete:
+            copied_user = TempSubscribersData(team_ids=row.team_ids, subscriber_emails=row.subscriber_emails)
+            subscribed_users.append(copied_user)
+
+        self.db_session.query(TempSubscribersData).delete(synchronize_session=False)
```
AndrewKozovyi, line 37: dal

database/alembic.ini
Avariq, line 64: do not commit

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,39 @@
...
+        new_rows = [
+            TempSubscribersData(team_ids=preference_index, subscriber_emails=user.email)
+            for user in users
+        ]
+
+        self.db_session.add_all(new_rows)
+        self.db_session.commit()
+
+    def __get_users_by_preference_index(self, preference_index):
+        query = (
```
Iarynovskyi, line 19: ...

service/implementation/email_sender/templates/news_notification_email.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Title</title>
+</head>
+ <body style="font-family: Arial, sans-serif; line-height: 1.6; color: #333; background-color: #f9f9f9; padding: 20px;">
+     <div style="max-width: 600px; margin: 0 auto; background: white; border: 1px solid #ddd; border-radius: 10px; padding: 20px;">
+         <h2>Dear, <strong>{{username}}! </strong> </h2>
+         <p> Your preference team <strong>{{team_name}}</strong> has appeared on the website right now!</p>
```
Avariq, line 10:
```suggestion
         <p> Your preference team <strong>{{team_name}}</strong> has appeared on the website just now!</p>
```

service/implementation/email_sender/templates/news_notification_email.html
```diff
@@ -0,0 +1,17 @@
...
+         <p> Be the first to read about their achievements </p>
```
Avariq, line 11:
```suggestion
         <p> Be the first to read about its achievements </p>
```

or

```suggestion
         <p> Be the first to read about it </p>
```

service/implementation/email_sender/test.py
```diff
@@ -0,0 +1,9 @@
+from api.routes.__init__ import app
+
+def test():
```
Avariq, line 3: Don't want to see any test/commented/legacy/etc code in PRs.
Please, have it removed

service/implementation/email_sender/user_subscription_manager.py
Avariq: Please tend to the comment above.

Avoid depending on file structure when you can.

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,52 @@
...
+    def try_add_subscribers_to_temp_table(self, team_index, news_name):
+        self.__user_subscription_dal.try_add_subscribers_data(team_index, news_name)
+
+    def __on_subscribers_inserted(self, mapper, connection, target):
+        self.__try_send_email_to_users()
+
+    def __try_send_email_to_users(self):
+        subscribed_users = self.__user_subscription_dal.pop_subscribed_users_data()
+
+        for user in subscribed_users:
```
Avariq, line 29: We should use some sort of Parallel Execution here.

The best approach would be to just use bulk_send emails but I am not sure whether Gmail supports that in the way we need (1. personalized emails 2. user A shouldn't know about user B getting the same email). So, yeah, at least we should make this loop run faster with threads

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,52 @@
...
+        with app.app_context():
+            msg = Message(
+                "News notification",
+                sender=current_app.config['MAIL_USERNAME'],
+                recipients=[user_email],
+                html=self.__get_email_template(news_name, username, team_name)
+            )
+            current_app.extensions['mail'].send(msg)
+
+            return CommonResponse().to_dict()
```
Avariq, line 43: what's that for?

service/implementation/email_sender/user_subscription_manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __get_email_template(self, news_name, username, team_name):
+        template_dir = os.path.join(os.path.dirname(__file__), "templates")
```
Avariq, line 46: We'd better store this path somewhere. 
I don't particularly like the idea of being dependent on constructing paths each time, because it makes code more coupled which goes against SOLID.

api/container/container.py
```diff
@@ -12,17 +13,18 @@
...
 
 
-
 class Container(containers.DeclarativeContainer):
     wiring_config = containers.WiringConfiguration(
         modules=[
             "api.routes.api_login",
             "api.routes.api_news",
             "api.routes.api_user_preferences",
+            "service.implementation.email_sender.user_subscription_manager",
```
Avariq, line 27: I'd move this after all API declarations (it's usually a good thing to sort names in A-Z order)

database/azure_blob_storage/save_get_blob.py
```diff
@@ -313,30 +313,32 @@ def save_news_index_to_db(blob_name: str, json_data,  session) -> None:
...
                 for name in team_name:
-                    team_index_id = team_dict.get(name, None)
-                    if team_index_id is not None:
-                        team_index = TeamInNews(
-                            news_id=news_index.news_id,
-                            name=name,
-                            team_index_id=team_index_id
-                        )
-                        session.add(team_index)
+                    __add_team_index_to_db(team_dict, name, news_index, blob_name, session)
```
Avariq, line 316: We have two identical lines.

You can do smth like:

`if not isinstance...:`
`    team_name = [team_name,]`
` `
`for name in team_name:`
`    your logic`

database/azure_blob_storage/save_get_blob.py
```diff
@@ -313,30 +313,32 @@ def save_news_index_to_db(blob_name: str, json_data,  session) -> None:
...
+                __add_team_index_to_db(team_dict, team_name, news_index, blob_name, session)
 
         session.commit()
     except Exception as e:
         session.rollback()
         print_error_message(f"Error when saving the news index in the database: {e}")
 
 
+def __add_team_index_to_db(team_dict, team_name, news_index, blob_name, session):
+    from api.container.container import Container
```
Avariq, line 327: Why here? Is there some sort of a circular import going on?

database/postgres/dal/user_subscription.py
```diff
@@ -0,0 +1,40 @@
...
+                team_ids=row[0].team_ids,
+                subscriber_emails=row[0].subscriber_emails,
+                news_name=row[0].news_name,
+                username=row[1]
+            )
+            subscribed_users_data_for_email.append(copied_user)
+
+        self.db_session.query(TempSubscribersData).delete(synchronize_session=False)
+
+        return subscribed_users_data_for_email
```
Avariq, line 40:
```suggestion
    def pop_subscribed_users_data(self):
        rows_to_delete = (
            self.db_session.query(TempSubscribersData, User.username)
            .join(User, TempSubscribersData.subscriber_emails == User.email)
            .join(TeamIndex, TempSubscribersData.team_ids == TeamIndex.team_index_id)
            .all()
        )

        subscribed_users_data_for_email = [
            TempSubscriberDataWithUsernameDto(
                team_ids=row[0].team_ids,
                subscriber_emails=row[0].subscriber_emails,
                news_name=row[0].news_name,
                username=row[1]
            )
            for row in rows_to_delete
        ]

        self.db_session.query(TempSubscribersData).delete(synchronize_session=False)

        return subscribed_users_data_for_email

```
How about this?

dto/api_output.py
```diff
@@ -85,6 +85,21 @@ class OutputRecommendationList(Schema):
     rating = fields.Int(required=True)
 
 
+class TempSubscriberDataWithUsernameDto(Schema):
```
Avariq, line 88: Is there a version "without username"? If not - please do:

```suggestion
class TempSubscriberDataDto(Schema):
```

Avariq: I also think we name all our dtos with capitals: `DTO`

not sure though

service/implementation/email_sender/user_subscription_manager.py
Avariq: Please either fix it (if it's easy) or jusr resolve this conversation and let it be as is

database/azure_blob_storage/save_get_blob.py
IceStorman: Yes, circular import is the only reason why this import is located here

> Why here? Is there some sort of a circular import going on?

PR #31, opened 2025-03-06
UJP-178   New save logic of Games

Iarynovskyi:
UJP-178


database/postgres/save_data.py
```diff
@@ -4,13 +4,25 @@
...
 FIXTURE = 'fixture'
 SPORT_TO_SAVE_TEAM_AS_LEAGUE = ['mma', 'formula-1']
 PLAYERS = ['fighters', 'drivers', 'players']
+CUSTOM_GAMES = {
+    'cancelled': ['Cancelled', 'Suspended'],
+    'finished': ['Finished', 'Awarded', 'After Over Time', 'Completed', 'After Extra Time', 'After Penalties', 'Completed'],
+    'abandoned': ['Abandoned', 'Interrupted'],
+    'postponed': ['Postponed'],
+    'scheduled': ['Not Started', 'Time To Be Defined', 'Scheduled'],
+    'not_played': ['Technical Loss', 'WalkOver', ],
```
AndrewKozovyi, line 21: no comma inside in the end needed, as we finished our list

database/postgres/save_data.py
```diff
@@ -87,8 +99,21 @@ def save_api_data(json_data: Dict, sport_name: str) -> None:
...
 
-                if not isinstance(status, str) and isinstance(status, dict):
-                    status = status.get('long')
+                if isinstance(status, dict):
+                    status = status.get('long', '')
+
+                if not isinstance(status, str) or not status:
+                    status = IN_PROGRESS
+
+                status_lower = status.lower()
```
Avariq, line 108:
```suggestion
                status_lower = status.lower()
                game_type_result = None
                
                for game_type, phrases in CUSTOM_GAMES.items():
                    if any(phrase in status_lower for phrase in map(str.lower, phrases)):  
                        game_type_result = game_type
                        break

```

1. Avoid using `type` in variable naming, since this word is reserved by python
2. Preprocess all phrases at once instead of doing .lower() for each separately

database/migrations/versions/300803741dca_new_column_types_for_games.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '300803741dca'
+down_revision: Union[str, None] = '75c86cc206ae'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('Games', sa.Column('type', sa.String(), nullable=True))
```
Avariq, line 23: Do we have a predefined set of these "types" we can store or is it some random stuff?

because if it's not random, I'd rather have you created separate DB table (e.g. game_statuses) to hold expected game statuses there.

And this table will just reference that table via FK and store type_id instead of string

Iarynovskyi: I do not see much point in this, because we already have so many calls to BD.
There are sooooo many in this pars-logic, so I think it's better without this separate DB table

database/migrations/versions/9a5daa74ec72_new_table_for_url.py
```diff
@@ -0,0 +1,38 @@
...
+revision: str = '9a5daa74ec72'
+down_revision: Union[str, None] = '8d79017f4a3e'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('StreamUrl',
+    sa.Column('stream_url_id', sa.Integer(), nullable=False),
```
Avariq, line 24: What's the purpose of this table?

Is it even a part of this PR?

database/migrations/versions/df7bd7ffe913_hai_bog_vrytue_this_migration.py
```diff
@@ -0,0 +1,57 @@
+"""hai_bog_vrytue_this_migration
```
Avariq, line 1: Got bit by Taras? What's that naming for?

database/models/streams.py
```diff
@@ -3,6 +3,6 @@
 class Stream(Base):
     __tablename__ = 'Streams'
     stream_id = Column(Integer, primary_key=True)
-    stream_url = Column(String) 
     start_time = Column(DateTime)
-    sport_id = Column(ForeignKey('Sports.sport_id'))
\ No newline at end of file
+    sport_id = Column(ForeignKey('Sports.sport_id'))
```
Avariq, line 7: Many changes for Streams here. Is it intended?

database/postgres/save_data.py
```diff
@@ -1,16 +1,31 @@
+from database.models import GamesStatuses
 from database.session import SessionLocal
 from typing import Dict
+from cachetools import TTLCache
 from database.postgres.dto import TeamDTO, CountryDTO, LeagueDTO, GameDTO, SportDTO, PlayerDTO
 from database.postgres.dal import TeamDAL, LeagueDAL, GameDAL, CountryDAL, SportDAL, PlayerDal
 from datetime import datetime
 
+status_cache = TTLCache(maxsize=10, ttl=60*60*24)
```
Avariq, line 9: MINOR: If that's 24 hours - too much time.

MINOR: I don't know exactly how this logic is used, since it is part of the scraper, but if there is a possibility of more than one thread running, we should probably make it Thread-Safe (see the sample image below). I am, however, 99% sure that we don't do multiple threads here. I think all of our scrapers are synchronous. 
image

database/migrations/versions/9a5daa74ec72_new_table_for_url.py
Avariq: Is this table's only purpose to store stream url separately? If that's the case then why? Can one stream have multiple urls?

Iarynovskyi: This table was created by Andriy for task with streams
So that there are no two heads again, we do all migrations here

database/migrations/versions/df7bd7ffe913_hai_bog_vrytue_this_migration.py
Iarynovskyi: I'm tired of the alembic errors..

database/models/streams.py
Iarynovskyi: Andriy changes again

database/postgres/save_data.py
```diff
@@ -1,16 +1,31 @@
...
 SPORT_TO_SAVE_TEAM_AS_LEAGUE = ['mma', 'formula-1']
 PLAYERS = ['fighters', 'drivers', 'players']
+CUSTOM_GAMES = {
+    'cancelled': ['Cancelled', 'Suspended'],
+    'finished': ['Finished', 'Awarded', 'After Over Time', 'Completed', 'After Extra Time', 'After Penalties', 'Completed'],
+    'abandoned': ['Abandoned', 'Interrupted'],
+    'postponed': ['Postponed'],
+    'scheduled': ['Not Started', 'Time To Be Defined', 'Scheduled'],
+    'not_played': ['Technical Loss', 'WalkOver'],
+
```
Taras4568, line 25: why empty lineeeeeeeeee??????? aaaaaaaaaaaaaa

PR #32, opened 2025-03-07
UJP-92 UJP-179 Streams save logic to db and streams endpoint

AndrewKozovyi:
UJP-92
UJP-179
I've made save logic to db and getting info from db, used Roman's filters for getting form db and made scraper to look for streams

!!!!I AM NOT RESPONSIBLE FOR TARAS'S ACTIONS ON THIS PR !!!!


database/postgres/dal/stream.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+
+        self.session.add(new_stream)
+        self.session.commit()
+
+        return new_stream
+
+
+    def create_stream_status(self, status_dto: StreamStatusDTO) -> Streams_Status:
+        new_stream_status = Streams_Status(
```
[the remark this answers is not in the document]
Avariq: I also don't get the idea of creating `Stream_Status` row with just a `stream_id`. Doesn't seem alright

database/postgres/dal/stream.py
```diff
@@ -0,0 +1,57 @@
...
+            stream_url=url_dto.stream_url,
+        )
+
+        self.session.add(new_url)
+        self.session.commit()
+
+        return new_url
+
+
+    def save_stream_statuses(self, status_dto_list: List[StreamStatusDTO]):
```
Avariq, line 50: Move this method above the `create_stream_url`

(statuses logic together and urls logic together)

database/postgres/dto/stream.py
```diff
@@ -0,0 +1,31 @@
...
+    title: Optional[str] = Field(None)
+    start_time: Optional[datetime] = Field(None)
+    sport_id: Optional[int] = Field(None)
+    stream_urls: Optional[list] = Field(None)
+
+    class Config:
+        from_attributes = True
+
+
+class StreamStatusDTO(BaseModel):
```
Avariq, line 16:
```suggestion
class StreamStatusDTO(BaseModel):
    streams_status_id: Optional[int] = None
    status_id: Optional[int] = None
    stream_id: Optional[int] = None

    model_config = {"from_attributes": True}
```

can't you just do this? (It's supposed to be pydantic v2 approach while yours is v1. Not sure though)

database/postgres/dto/stream.py
```diff
@@ -0,0 +1,31 @@
+from datetime import datetime
+from pydantic import BaseModel, Field
+from typing import Optional
+
+class StreamDTO(BaseModel):
+    stream_id: Optional[int] = Field(None)
```
Avariq, line 6: Is it alright to have all of the fields Optional?

dto/api_output.py
```diff
@@ -27,6 +27,14 @@ class TeamsLeagueOutputWithCount(Schema):
     teams = fields.List(fields.Nested(TeamsLeagueOutput))
     count = fields.Int()
 
+class StreamsOutput(Schema):
```
Avariq, line 30: This one is marshmellow, right?

service/api_logic/streams_logic.py
```diff
@@ -1,126 +1,48 @@
...
-@handle_exceptions
-@logger.log_function_call()
-def get_streams_of_sport(session, sport_name:str):
-    sport = get_sport_index_by_name(session, sport_name)
-    filters = [Stream.sport_id == sport.sport_id]
-    streams = fetch_streams(session, order_by=Stream.start_time.desc(), filters=filters)
-    return json_streams(streams)
+        streams = self._stream_dal.query_output(filtered_query)
+        stream_output = StreamsOutput(many=True)
+        streams = stream_output.dump(streams)
```
Avariq, line 32: We can just do that in one line:

```suggestion
        stream_output = StreamsOutput(many=True).dump(streams)
```

service/api_logic/streams_logic.py
```diff
@@ -1,126 +1,48 @@
...
 
+        response_dto = ListResponseDTO()
 
+        return response_dto.dump({"items": streams, "count": count})
```
Avariq, line 36: I hate those hardcoded strings. Makes me wanna pull your project and try to fix that myself somehow. There must be a way

service/api_logic/streams_logic.py
```diff
@@ -1,126 +1,48 @@
...
-            "start_time":stream.start_time,
-            "sport_id":stream.sport_id
-        })
-    return json.dumps(streams_data, ensure_ascii=False)
 
+    def save_streams_to_streams_table(self, streams_data):
+        for stream in streams_data:
+            if stream.stream_urls:
+                new_stream = self._stream_dal.create_stream(stream)
+                for url in stream.stream_urls:
```
Avariq, line 43: What if we manage to create a stream but fail creating links to it? (Some DB error, let's say)
Is it okay then for that stream to exist in our DB even though there are no links stored for it? 

If not (stream and urls both have to be added successfully) -> we need a transaction here, which will only commit the changes if everything has been added

streams_scraper/streams_scraper.py
```diff
@@ -1,27 +1,35 @@
...
     options.add_argument("--disable-dev-shm-usage")
     options.add_argument("--disable-extensions")
-    
-    service = Service(ChromeDriverManager().install())
-    driver = webdriver.Chrome(service=service, options=options)
-    
+    options.add_argument("--disable-blink-features=AutomationControlled")
+
+    user_agents = [
+        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
```
Avariq, line 26: MINOR:

There is actually a lib for generating user agents

streams_scraper/streams_scraper.py
```diff
@@ -35,84 +43,88 @@ def configure_driver():
...
     try:
-        
-        
-        games_today = get_games_today(session)
-        future_games = filter_future_games(games_today)
-
-        
-        game_data = search_game_links(driver, future_games, games_today, session)
+        dto = SearchDTO().load({
+            "filters": [
```
Avariq, line 102: There must be a way of creating these objects without this string-dancing shit...

dto/api_output.py
AndrewKozovyi: yes

service/api_logic/streams_logic.py
AndrewKozovyi: You mean to make this as 1 query?

PR #36, opened 2025-03-12
UJP-162 Feature/interactions with news 

torii044:
UJP-162


api/routes/api_interactions.py
```diff
@@ -0,0 +1,44 @@
...
+    response = {"error in data base": str(e)}
+    return response
+
+@interactions_app.route('/save', methods=['POST'])
+@logger.log_function_call()
+def save_interaction():
+    try:
+        data = request.get_json()
+        dto = InteractionsDTO().load(data)
+        save_interaction(dto)
```
Iarynovskyi, line 24: now we use services so make your own or move this to news_service

api/routes/api_interactions.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return {"message": "Interaction added successfully"}, 201
```
Iarynovskyi, line 26: hardcode and its not important info in front

api/routes/api_interactions.py
```diff
@@ -0,0 +1,44 @@
...
+    except CustomQSportException as e:
+        logger.error(f"Error in POST /: {str(e)}")
+        get_custom_error_response(e)
```
Iarynovskyi, line 29: here must be return

api/routes/api_interactions.py
```diff
@@ -0,0 +1,44 @@
...
+
+@interactions_app.route('/getStatus', methods=['GET'])
+@logger.log_function_call()
+def get_interaction_status_by_user_id():
+    try:
+        user_id = request.args.get("user_id")
```
Iarynovskyi, line 35: maybe better use POST method to nor write this hardcode get

api/routes/api_interactions.py
```diff
@@ -0,0 +1,44 @@
...
+        news_id = request.args.get("news_id")
+        interaction_type = request.args.get("interaction_type")
+        dto = InteractionsDTO.load({"news_id": news_id, "user_id": user_id, "interaction_type": interaction_type})
```
Iarynovskyi, line 38: and here with method POST u can use only .load(data)

api/routes/api_interactions.py
```diff
@@ -0,0 +1,44 @@
...
+        status = get_interaction_status(dto)
+
+        return {"status": status}, 200
+    except CustomQSportException as e:
+        logger.error(f"Error in GET /: {str(e)}")
+        get_custom_error_response(e)
```
Iarynovskyi, line 44: return

database/migrations/versions/0d93e8ea82b6_add_initial_values_to_interactions.py
```diff
@@ -0,0 +1,44 @@
+"""add initial values to interactions
```
Iarynovskyi, line 1: its bad...
we create all migrations in data-base-architecture branch
now we have conflicts in migrations and 2 heads

database/models/__init__.py
```diff
@@ -10,14 +10,15 @@
 from .League import League
 from .game import Games
 from .teams_in_news import TeamInNews
-from .likes import Likes
-from .views import Views
```
Iarynovskyi, line 14: maybe now we have very big problem with conflict in migrations..

database/models/news.py
```diff
@@ -7,3 +7,5 @@ class News(Base):
     sport_id = Column(ForeignKey('Sports.sport_id'))
     interest_rate = Column(Integer)
```
Iarynovskyi, line 8: my bad. I create this for likes but with bad naming, so soon will delete this

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,61 @@
...
+        setattr(interaction, 'timestamp', interaction_dto.timestamp)
+        setattr(interaction, 'interaction_type', interaction_type_id)
+        self.db_session.commit()
+        self.db_session.refresh(interaction)
+        return interaction
+
+    def get_interaction_by_user_id_and_type(self, user_id: int, interaction_type_id: int) -> Optional[InteractionWithNews]:
+        return self.db_session.query(InteractionWithNews).filter_by(user_id=user_id).filter_by(interaction_type=interaction_type_id).first()
+
+    def get_interaction_by_user_id_and_news_id_and_type(self, user_id: int, news_id: int, interaction_type_id: int) -> Optional[InteractionWithNews]:
```
Iarynovskyi, line 49: many 'and' here? maybe u can divide this to more func ?

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,51 @@
...
+    interaction_type_id = get_interaction_type_id(interaction_dto.interaction_type)
+    if interaction_type_id == InteractionTypes.UNDEFINED.value:
+        raise UndefinedInteractionType(interaction_dto.interaction_type)
+
+    interaction_entry = dal.get_interaction_by_user_id_and_news_id_and_type(interaction_dto.user_id,
+                                                                            interaction_dto.news_id,
+                                                                            interaction_type_id)
+    if interaction_entry:
+        return True
+    return False
```
Iarynovskyi, line 51: maybe better to create some exception for False status ?  Because API will send to front status 200, but with test 'False'

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,51 @@
...
+def get_interaction_status(interaction_dto) -> bool:
+    interaction_type_id = get_interaction_type_id(interaction_dto.interaction_type)
+    if interaction_type_id == InteractionTypes.UNDEFINED.value:
+        raise UndefinedInteractionType(interaction_dto.interaction_type)
+
+    interaction_entry = dal.get_interaction_by_user_id_and_news_id_and_type(interaction_dto.user_id,
+                                                                            interaction_dto.news_id,
+                                                                            interaction_type_id)
+    if interaction_entry:
+        return True
```
Iarynovskyi, line 50: And this information is useless in front

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,57 @@
...
+        :param news_id: The ID of the article.
+        :param interaction_type_id: The type of interaction.
+        :return: An InteractionWithNews object if found, otherwise None.
+        """
+        return self.db_session.query(InteractionWithNews).filter_by(user_id=user_id).filter_by(news_id=news_id).filter_by(interaction_type=interaction_type_id).first()
+
+    def get_interaction_by_id(self, interaction_id: int) -> Optional[InteractionWithNews]:
+        return self.db_session.query(InteractionWithNews).filter_by(interaction_id=interaction_id).first()
+
+    def delete_interaction(self, interaction_id: int) -> bool:
```
Taras4568, line 51: imo it will be better to write something like: 
if interaction:
        self.db_session.delete(interaction)
        self.db_session.commit()

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,63 @@
...
+                timestamp=interaction_input_dto.timestamp,
+            )
+
+            return db_dto
+
+        except KeyError:
+            raise IncorrectInteractionType(interaction_input_dto.interaction_type)
+
+
+    def save_interaction(self, interaction_input_dto):
```
Taras4568, line 38: to much hardcoding, should be rethinked

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,63 @@
+from exept.exeptions import IncorrectInteractionType, BlobFetchError, DatabaseConnectionError
+from service.api_logic.models.api_models import InteractionTypes
+from database.postgres.dto import InteractionWithNewsDTO
+from logger.logger import Logger
+
+LIKE_ID = InteractionTypes['LIKE'].value
+DISLIKE_ID = InteractionTypes['DISLIKE'].value
+
+class InteractionWithNewsService:
+
```
Taras4568, line 10: redundant empty line

database/postgres/dal/interaction_with_news.py
torii044: all my DALs are written like this. i'm just trying to be consistent

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,63 @@
...
+            news_entry = self._news_dal.get_news_by_id(interaction_input_dto.blob_id)
+
+            if not news_entry:
+                raise BlobFetchError(interaction_input_dto.blob_id)
+
+            db_dto = InteractionWithNewsDTO(
+                news_id=news_entry.news_id,
+                user_id=interaction_input_dto.user_id,
+                interaction_type=interaction_type_id,
+                timestamp=interaction_input_dto.timestamp,
```
Iarynovskyi, line 29: u use here timestamp from front but better do this in back by datetime.now() or smth like this
becouse React cool framework but it use users local time and it can be different to many users

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,57 @@
...
+            interaction_type=interaction_dto.interaction_type,
+            timestamp=interaction_dto.timestamp
+        )
+        self.db_session.add(new_interaction)
+        self.db_session.commit()
+        return new_interaction
+
+    def update_interaction(self, interaction_id: int, interaction_dto) -> Optional[InteractionWithNews]:
+        interaction = self.get_interaction_by_id(interaction_id)
+        if not interaction:
```
Taras4568, line 29: Pointless if
you can write something like:
def update_interaction(self, interaction_id: int, interaction_dto) -> Optional[InteractionWithNews]:
    interaction = self.get_interaction_by_id(interaction_id)
    
    if interaction:
        setattr(interaction, 'timestamp', interaction_dto.timestamp)
        setattr(interaction, 'interaction_type', interaction_dto.interaction_type)
        self.db_session.commit()
        self.db_session.refresh(interaction)
    
    return interaction  
and function will return None by default if interaction condition will not be fulfilled

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -42,6 +43,132 @@ export default function InsideNewsPage() {
...
+    const saveInteraction = async (interactionType) => {
+        if (user) {
+            try {
+                await axios.post(
+                    `${apiEndpoints.url}${apiEndpoints.interactions.saveInteraction}`,
+                    {
+                        user_id: user.id,
+                        blob_id: articleId,
+                        interaction_type: interactionType,
+                        timestamp: new Date().toISOString(),
```
Iarynovskyi, line 123: yes u use 'toISOString()' here and its get users local time sooo..  this better do in back that i wrote before or after, idk how work this PR system ^)

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,57 @@
...
+            news_id=interaction_dto.news_id,
+            user_id=interaction_dto.user_id,
+            interaction_type=interaction_dto.interaction_type,
+            timestamp=interaction_dto.timestamp
+        )
+        self.db_session.add(new_interaction)
+        self.db_session.commit()
+        return new_interaction
+
+    def update_interaction(self, interaction_id: int, interaction_dto) -> Optional[InteractionWithNews]:
```
Iarynovskyi, line 27: u can write one method for updating interaction status by using this ORM
I think its could be better

Iarynovskyi (changes requested):

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,57 @@
...
+    def get_interaction(self, user_id: int, news_id: int, interaction_type_id: int) -> Optional[InteractionWithNews]:
+        """
+        Retrieves a user's interaction with a news item based on given parameters.
+
+        :param user_id: The ID of the user.
+        :param news_id: The ID of the article.
+        :param interaction_type_id: The type of interaction.
+        :return: An InteractionWithNews object if found, otherwise None.
+        """
+        return self.db_session.query(InteractionWithNews).filter_by(user_id=user_id).filter_by(news_id=news_id).filter_by(interaction_type=interaction_type_id).first()
```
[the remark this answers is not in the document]
Iarynovskyi: `.filter_by(InteractionWithNews. user_id == user_id, ...)`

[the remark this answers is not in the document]
Iarynovskyi:
```suggestion
        return self.db_session.query(InteractionWithNews).filter_by(InteractionWithNews.user_id == user_id, ... ).first()
```

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,45 @@
+from database.models import InteractionWithNews
```
Iarynovskyi, line 1: This comment i write for all logic of likes
Now i test this mechanic in your branch and when i go to news page i saw 10 likes with empty heart 
Than i liked it ->  and saw 11 likes with red heart
Than i live this page and at this moment front saved my reaction
Than i go back and saw 10 with red heart
Than i click on it ->   and saw 9 likes with empty heart

I think this logic must working not like this
And if again go to this page i saw 10 likes with empty heart
I think its minor but we could have -1 likes with this feature

torii044: now actual information about likes is being displayed. but i'm not sure if i'm doing it in an optimal way

database/postgres/dal/interaction_with_news.py
```diff
@@ -0,0 +1,45 @@
...
+        Retrieves a user's interaction with a news item based on given parameters.
+
+        :param user_id: The ID of the user.
+        :param news_id: The ID of the article.
+        :param interaction_type_id: The type of interaction.
+        :return: An InteractionWithNews object if found, otherwise None.
+        """
+        return self.db_session.query(InteractionWithNews).filter_by(user_id=user_id, news_id=news_id, interaction_type=interaction_type_id).first()
+
+    def get_interaction_by_id(self, interaction_id: int) -> Optional[InteractionWithNews]:
```
Avariq, line 44: I don't see it being used anywhere

Also, I don't think that'd be a good idea to get interaction based on a single id column. It's not exactly the best example, since our project `is unlikely to be getting such a big number of interactions, but:

I presume the this Id is Integer, which would result to be around couple of billion interactions MAX. Imagine having thousands of news with hundreds of thousands of like and views each day (we have some Instagram posts with millions of likes daily, for example) - we would run out of integer capacity quite soon that way.

The easy solution - just using bigint (which has 64bit instead of 32) but the correct one is:

Can we have an interaction with nothing? No, we cannot. Therefore we can create cluster keys (e.g interaction id + news_id) and then use them both each time to get the data required. This way we are safe in terms of integer overflow and our DB is normilized as well.

BTW:
Ordering by cluster keys would also be faster in this case. It will also save some disk space for us

Summary:
This method should be either removed or restructured for more proficient use of integer memory

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -34,14 +37,156 @@ export default function InsideNewsPage() {
...
+    const initialLikeStatusRef = useRef(initialLikeStatus);
+
+    const { user } = useContext(AuthContext)
+
+
+    useEffect(() => {
+        const observer = new IntersectionObserver(
+            ([entry]) => {
+                if (entry.isIntersecting && !hasRead) {
+                    saveInteraction('read');
```
Avariq, line 61: I don't like hardcoded strings like "read" here

Avoid that if possible.
For example you can create function `saveReadInteraction` and hardcode the string there (if that would be the only place it is used)

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -34,14 +37,156 @@ export default function InsideNewsPage() {
...
+                await axios.post(
+                    `${apiEndpoints.url}${apiEndpoints.interactions.saveInteraction}`,
+                    {
+                        user_id: user.id,
+                        blob_id: articleId,
+                        interaction_type: interactionType,
+                    }
+                );
+            } catch (error) {
+                toast.error(`Troubles with saving interaction: ${error}`)
```
Avariq, line 127: "Failed to save" sounds better IMO

Do that to all errors (even better to ask GPT-guy about the best error message)

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -34,14 +37,156 @@ export default function InsideNewsPage() {
...
+    const handleLikeStatus = async () => {
+        if (likeStatusRef.current !== initialLikeStatusRef.current) {
+            let interactionType = likeStatusRef.current ? 'like' : 'dislike';
+            saveInteraction(interactionType);
+        }
+    };
+
+    useEffect(() => {
+        const logInteraction = async () => {
+            await saveInteraction('open');
```
Avariq, line 110: You hardcode a lot of those interactionType strings. Maybe create some enum-like stuff. Though JS does not have exactly an ENUM type, but you can work it around like:

`const InteractionType = Object.freeze({`
`    LIKE: "like",`
`    DISLIKE: "dislike",`
`    COMMENT: "comment",`
`    SHARE: "share"`
});`

Something of that kind

Avoid hardcode at all cost. We have to minimize room for error in the future

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -34,14 +37,156 @@ export default function InsideNewsPage() {
...
+            }
+        }
+    };
+
+    useEffect(() => {
+        async function getInteractionsCounts() {
+            try {
+                const { data } = await axios.get(`${apiEndpoints.url}${apiEndpoints.interactions.getInteractionsCounts}`, {
+                    params: {
+                        blob_id: articleId,
```
Avariq, line 137: I don't particularly like `blob_id`. I do realize what that means, but since we only store articles in our blob (as far as I know that is) maybe we could go with something like "article_blob_id" at least?

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -34,14 +37,156 @@ export default function InsideNewsPage() {
...
+    const toggleLike = () => {
+        if (user) {
+            setLikeStatus(prev => {
+                const newStatus = !prev;
+                setLikes(count => newStatus ? count + 1 : count - 1);
+                return newStatus;
+            });
+        }
+        else {
+            const notify = () => toast('Sign in to leave your reaction', {
```
Avariq, line 181:
```suggestion
            const notify = () => toast('Sign in to like this post', {
```

Something like that maybe

We should always assume that our user is a 90-year-old grandma who just got her phone yesterday -> Everything should be as plain and understandable as humanly possible

frontend/src/pages/news/insideNewsPage.jsx
```diff
@@ -34,14 +37,156 @@ export default function InsideNewsPage() {
...
+                action: {
+                    label: 'sign in',
+                    onClick: () =>  navigate('/sign-in') ,
```
Avariq, line 184: I would also like to see this "/sign-in" specified somewhere in some sort of a variable or whatever.

Imagine our project grows and has dozens of places with login page readdress. And then imagine we need to rename it from sign-in to something else. It's much better to have one place to change than 10+, is it not?

Talk this through with The Boys to decide on the best place to put that stuff

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+from datetime import datetime
+
+LIKE_ID = InteractionTypes['LIKE'].value
+DISLIKE_ID = InteractionTypes['DISLIKE'].value
+
+class InteractionWithNewsService:
+    def __init__(self, interaction_with_news_dal, news_dal):
+        self._interaction_with_news_dal = interaction_with_news_dal
+        self._logger = Logger("logger", "all.log").logger
+        self._news_dal = news_dal
```
Avariq, line 15: Як казав мій лектор з вишмату: "Мухи до мух, а котлєти до котлєт" ->

Дали до далів, логери до логерів (і відділи пустим рядком ще)

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,63 @@
...
+            if interaction_entry:
+                self._interaction_with_news_dal.update_interaction(interaction_entry.interaction_id, interaction_dto)
+            else:
+                self._interaction_with_news_dal.save_interaction(interaction_dto)
+
+        elif interaction_dto.interaction_type == DISLIKE_ID:
+            interaction_entry = self._interaction_with_news_dal.get_interaction(interaction_dto.user_id, interaction_dto.news_id, LIKE_ID)
+
+            if interaction_entry:
+                self._interaction_with_news_dal.update_interaction(interaction_entry.interaction_id, interaction_dto)
```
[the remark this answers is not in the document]
Avariq: I agree

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+            db_dto = InteractionWithNewsDTO(
+                news_id=news_entry.news_id,
+                user_id=interaction_input_dto.user_id,
+                interaction_type=interaction_type_id,
+                timestamp=datetime.now()
+            )
+
+            return db_dto
+
+        except KeyError:
```
Avariq, line 35: If that KeyError is focused on `blob_id` alone, it'd be better to just check that `blob_id` explicitly (some `if` statement or whatever) than catching an exception. Cleaner approach and better performance as well.

Basically:

It's better to do `if (true)` -> raise Exception
than `try ... except ... raise`

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+
+            db_dto = InteractionWithNewsDTO(
+                news_id=news_entry.news_id,
+                user_id=interaction_input_dto.user_id,
+                interaction_type=interaction_type_id,
+                timestamp=datetime.now()
+            )
+
+            return db_dto
+
```
Avariq, line 34: MINOR:
This empty line before `except` is redundant IMO

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+                return
+
+        self._interaction_with_news_dal.save_interaction(interaction_dto)
+
+
+    def get_interaction_status(self, interaction_input_dto) -> bool:
+        interaction_dto = self.__convert_input_to_db_dto(interaction_input_dto)
+        interaction_entry = self._interaction_with_news_dal.get_interaction(interaction_dto.user_id, interaction_dto.news_id, interaction_dto.interaction_type)
+
+        return True if interaction_entry else False
```
Avariq, line 63: can't we just do

```suggestion
        return bool(interaction_entry)
```

?

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+
+            if interaction_entry:
+                self._interaction_with_news_dal.update_interaction(interaction_entry.interaction_id, interaction_dto)
+                return
+
+        self._interaction_with_news_dal.save_interaction(interaction_dto)
+
+
+    def get_interaction_status(self, interaction_input_dto) -> bool:
```
Avariq, line 59: `get_interaction_status` naming is a little vague. Can we extend the name a little to make it easier to understand (like: what exactly that "status" is/means?)

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+        interaction_entry = self._interaction_with_news_dal.get_interaction(interaction_dto.user_id, interaction_dto.news_id, interaction_dto.interaction_type)
+
+        return True if interaction_entry else False
+
+
+    def get_interactions_counts(self, interaction_input_dto):
+        interactions_counts = self._news_dal.get_interaction_counts(interaction_input_dto.blob_id)
+
+        if not interactions_counts:
+            raise BlobFetchError(interaction_input_dto.blob_id)
```
Avariq, line 70: I am not very sure about this exception (and a couple places in this PR before this one as well).

We fetch and store all our likes/dislikes/whatever in OUR database (not in blob, right? Plus - if user can like an article it means we already managed to fetch it from our Blob storage and display it on the page which makes this exception even more questionable), so I'd expect some "local" exception here e.g. `NewsNotFoundError`. Something of that sort.

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,75 @@
...
+
+        return OutputInteractions().dump({
```
Avariq, line 72:
```suggestion
        return OutputInteractions(likes=interactions_counts[0], views=interactions_counts[1]).model_dump()
```

I wonder if we can do something like this instead

service/api_logic/models/api_models.py
```diff
@@ -26,3 +26,11 @@ def __init__(self):
         self.related_name = "name"
         self.related_logo = "logo"
         self.related_id = "team_index_id"
+
+
+class InteractionTypes(Enum):
```
Avariq, line 31: We should have the exact same thing on our frontend (to avoid hardcoding strings like I mentioned above). Maybe make a cachable API call to retrieve this to frontend

Avariq (commented):
The only way to understand whether your .js code works fine is to test it I'm afraid. Don't see anything particularly sus, but please test it properly or maybe do cross testing (other teammemebrs test your code while you test theirs)

service/api_logic/interactions_logic.py
Iarynovskyi: А вона на Тесак випадково ?
Бо у нас теж є Мухи окремо, котлети окремо

database/postgres/dal/news.py
```diff
@@ -151,6 +151,8 @@ def delete_dataframe(self, dataframe):
         del dataframe
 
 
+    def get_interaction_counts(self, blob_id: str):
```
Avariq, line 154: Please make sure that we don't spam our DB with this call.

This method is useful in cases when we work with one News only, but if we load a page with, let's say, 20 News and each of those has some like and dislike number that has to be displayed. If we make 20 calls for such a case - it should be reimagined.

We already "draw" our news somewhere. News DB schema already has likes and dislikes so we just have to work with that.

Basically:
When we display our news on some page we should get both: the news info and likes amount with one DB call (which, again, is easy, since you properly designed the DB schema)

You should probably discuss this with Roman, since he is working with pagination

Hope that all made some sense :)
If not - will discuss during the meeting or just DM me in Telegram

frontend/src/interactionContext.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+const InteractionContext = createContext();
+
+export const InteractionProvider = ({ children }) => {
+    const [interactionTypes, setInteractionTypes] = useState(() => {
+        return JSON.parse(localStorage.getItem('interactionTypes')) || {};
+    });
+
+    useEffect(() => {
+        const fetchInteractionTypes = async () => {
```
Avariq, line 14:
```suggestion
        const getInteractionTypes = async () => {
```

frontend/src/pages/news/insideNewsPage.jsx
Avariq: We discussed it on one of our meetings: those links (e.g. sign-up in our case) should be defined somewhere (in one place). Ask the guys. I believe they have already put it somewhere

api/routes/api_interactions.py
```diff
@@ -0,0 +1,81 @@
...
+    except CustomQSportException as e:
+        logger.error(f"Error in GET /: {str(e)}")
+        return get_custom_error_response(e)
+
+
+@interactions_app.route('/fetchInteractionTypes', methods=['GET'])
+@logger.log_function_call()
+@inject
+@handle_exceptions
+def fetch_interaction_types():
```
Avariq, line 74: I'd also add caching here. InteractionTypes are unlikely to change frequently (maybe 30 minutes? something like that, idk. up to you)

api/routes/api_interactions.py
```diff
@@ -0,0 +1,81 @@
...
+        interactions_counts = service.get_interactions_counts(dto)
+
+        return interactions_counts
+
+    except CustomQSportException as e:
+        logger.error(f"Error in GET /: {str(e)}")
+        return get_custom_error_response(e)
+
+
+@interactions_app.route('/fetchInteractionTypes', methods=['GET'])
```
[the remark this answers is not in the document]
AndrewKozovyi: shouldn't it just be interactionTypes, because due to REST route shouldn't say what it does, am I wrong?

frontend/src/interactionContext.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { createContext, useContext, useEffect, useState } from 'react';
+import apiEndpoints from "./apiEndpoints";
+import axios from 'axios';
+import {toast} from "sonner";
+
+const InteractionContext = createContext();
+
+export const InteractionProvider = ({ children }) => {
```
AndrewKozovyi, line 8: may I be wrong, nbut isn't it better to store sypes as some "global variables" instead of sending a call to db every hour?

frontend/src/js/App.js
```diff
@@ -41,70 +42,72 @@ function App() {
     return (
         <>
             <TranslationsProvider>
-                <Router
-                    future={{
-                        v7_startTransition: true,
-                        v7_relativeSplatPath: true,
-                    }}>
+                <InteractionProvider>
```
AndrewKozovyi, line 45: if u still use this context in future, wrap it in index.js file for better visual look

frontend/src/interactionContext.jsx
```diff
@@ -0,0 +1,39 @@
...
+    const [interactionTypes, setInteractionTypes] = useState(() => {
+        return JSON.parse(localStorage.getItem('interactionTypes')) || {};
+    });
+
+    useEffect(() => {
+        const getInteractionTypes = async () => {
+            try {
+                const response = await axios.get(`${apiEndpoints.url}${apiEndpoints.interactions.getTypes}`);
+                const types = response.data;
+                localStorage.setItem('interactionTypes', JSON.stringify(types));
```
Avariq, line 18: MINOR:

I do agree with Andrew on this: it's not the best idea to hold these inside local storage. I would probably go for caching response on our API endpoint if I were you. Or maybe some dynamic JS varibale.

BUT this is alright. Leave it be

api/routes/api_interactions.py
```diff
@@ -0,0 +1,82 @@
+from dependency_injector.wiring import Provide, inject
```
Taras4568, line 1: Screenshot 2025-04-07 213004
 __init__ doesn't start

service/api_logic/interactions_logic.py
```diff
@@ -0,0 +1,56 @@
...
+DISLIKE_ID = InteractionTypes['DISLIKE'].value
+
+class InteractionWithNewsService:
+    def __init__(self, interaction_with_news_dal, news_dal):
+        self._interaction_with_news_dal = interaction_with_news_dal
+        self._news_dal = news_dal
+
+        self._logger = Logger("logger", "all.log").logger
+
+    def __convert_input_to_db_dto(self, interaction_input_dto):
```
Avariq, line 18: MINOR:

You should always try to specify the parameter's type. That'll semplify both code production and readability.

(e.g.  param: string, param2: int)

Avariq (approved):

PR #40, opened 2025-03-20
UJP-172 UJP-185 connect leagues front to back

Iarynovskyi:
Main logic with getting info from back work here
Pagination work
Build logic for generating filters blocks 
But big problems with adjust this page. I did it but with hardcode... 

For finish i need to make filters block adaptation


frontend/src/pages/sport items/leaguePage.js
```diff
@@ -113,18 +157,57 @@ function LeaguePage() {
...
+    const [selectedModel, setSelectedModel] = useState("leagues");
+    const [filters, setFilters] = useState([]);
+    const [burgerMenu, setBurgerMenu] = useState(false)
+    const [menuIsOpen, setMenuIsOpen] = useState(false)
+    const [menuIcon, setMenuIcon] = useState(initialIcon)
+
+    useEffect(() => {
+
+        const handleResize = () => {
+            const smallScreen = window.innerWidth <= 1050
```
Avariq, line 171: MINOR:

Maybe we should also have this one moved somewhere.

That'd allow us to make sort of a "standard" to follow (all supported screens, for example)

frontend/src/pages/sport items/leaguePage.js
```diff
@@ -113,18 +157,57 @@ function LeaguePage() {
...
+            setBurgerMenu(smallScreen)
+        }
+
+        handleResize();
+        window.addEventListener("resize", handleResize);
+    }, []);
+
+    const handleOpenMenu = () => {
+        setMenuIsOpen(prev => !prev)
+        setMenuIcon(!menuIsOpen ? <FaTimes size={28} color="black" /> : initialIcon)
```
Avariq, line 181: MINOIR: 

Maybe we should keep all our colors and fonts (as well as their size) in some place?

We usually do that when we want to maintain proper color and font styles

Avariq (commented):
Please test UI-related changes on different PCs and monitors, so that we don't get "works fine on my computer" kind of lines later

PR #43, opened 2025-03-23
UJP-200 autogenerating swagger for user and sport controlers

Iarynovskyi:


Avariq (approved):
Looks OK

Pity we didn't notice this could be done earlier

PR #44, opened 2025-03-23
UJP-199 Unit tests for save data players saving

AndrewKozovyi:


unit_tests/test_save_data/test_save_players.py
```diff
@@ -0,0 +1,101 @@
...
+                            "id": 55,
+                            "name": "Beta Warriors"
+                        }
+                    ]
+                }
+            ]
+        }
+
+    def test_parameters_missing(self):
+        save_api_data(self.json_case_empty, self.sport_name)
```
Avariq, line 84: This is Integration Testing which will impact the database.
What I need is UnitTesting.

For example:
You test some Service (let's say UserService)

That service has its methods (e.g. create_user). Those methods, however depend on different DALs and other stuff.

Your job is to create Mocks of those dependencies. For example -> fake dal that will not actually call the database, instead you will just tell it "hey, when I give you this set of data, you return this stuff, ok?". And then you just test the "create_user" method itself. We can talk more on it during the meeting, ot you can just watch some videos on YouTube, because this one is a little to complicated for me to explain in a single GitHub comment

unit_tests/save_data_tests/test_save_players.py
```diff
@@ -0,0 +1,88 @@
...
+                {
+                    "id": 3,
+                    "name": "John Doe",
+                    "photo": "https://example.com/john_doe.png",
+                }
+            ]
+        }
+
+        save_api_data(json_case_full_info, self.sport_name)
+        mock_save_players.assert_called_once()
```
Avariq, line 53: Is there a way for us to also verify correct data being stored in fake DB?
"CalledOnce" is a very easy pass.

You can either try to test the data stired in the fake DB (would be nice, even though it's closer to integration testing) or check that `database.postgres.dal.PlayerDal.save_players` is invoked with a PSECIFIC set of parameters.

For example you expect team_index_id to equal 5 or smth and you expect the icon to be something. That kind of stuff. This way you still do UNIT testing, but in a proper way.

Basically, you make sure your `save_api_data` method processed info correctly and passed it to `database.postgres.dal.PlayerDal.save_players` in a proper way as well

Avariq (approved):
Will do
Approved

PR #45, opened 2025-03-25
Feature/ujp 170 streams connect to front

AndrewKozovyi:
I've connected back streams to front, waiting for another pr about streams to clsoe for this one


Avariq (approved):
Only JS here and it seems clean. Hope it works :)

PR #47, opened 2025-03-27
UJP-181 slider adjust in main page but not recent games block

Iarynovskyi:


frontend/src/scss/games.scss
```diff
@@ -1,9 +1,175 @@
...
+        overscroll-behavior: contain;
+        touch-action: pan-x;
+
+        &::-webkit-scrollbar {
+            display: none;
+            width: 0;
+            height: 0;
+        }
+
+        //.sport-options-controls {
```
Avariq, line 39: Is this sort of a prep for next PR or smth? If so, let it be, if not - remove

Avariq (approved):
Approved if this one behaves properly with different screens (no "flying" blocks of UI all over the place)

PR #48, opened 2025-03-27
UJP-174 connect front to back paginated news

AndrewKozovyi:
I've user filters for this to get pagination and order


service/api_logic/news_logic.py
```diff
@@ -70,7 +71,11 @@ def get_filtered_news(self, filters: NewsDTO):
         filtered_query, count = FilterManagerStrategy.apply_filters(News, query, filters)
 
         news = self._news_dal.query_output(filtered_query)
-        return self.json_news(news)
+        news_output = self.json_news(news).json
+
+        response_dto = ListResponseDTO()
+
+        return response_dto.dump({"items": news_output, "count": count})
```
Avariq, line 78: Do this instead

`return ListResponseDTO().dump({"items": news_output, "count": count})`

Avariq (approved):
Seems alright

It'd be awesome, btw, if you guys put screenshots of your UI changes now that I think about it

service/api_logic/news_logic.py
AndrewKozovyi: no, we have a model now, I've done it on streamas, and as it is merged now< i can do it even smoother)

AndrewKozovyi: it will look more like this:

`teams_output = TeamsLeagueOutput(many=True).dump(teams)
        response_dto = ListResponseDTO(items = teams_output, count = count)`

PR #50, opened 2025-03-28
UJP-176 fix recommendations after jwt

Iarynovskyi:


api/routes/api_user_preferences.py
```diff
@@ -34,6 +33,7 @@ def sport_preferences_endpoint(service: UserService = Provide[Container.user_ser
 
         if request.method == "POST":
             service.add_preferences(dto)
+            delete_recommendation_key(dto.user_id)
```
Avariq, line 36: MINOR:

Just for future: It'd be better to do some class, which would handle those caching actions and keep an eye on our TTL. Using plain functions like this is a little bit against OOP, but I'll allow it since we don't have much time left

Avariq (approved):
Seems alright

PR #52, opened 2025-03-29
Feature/ujp 203 unit tests for auth bl no jwt logic

AndrewKozovyi:
ujp-203 
I've made unit test for main logic of auth, without jwt logic (firstly, I'm scared of Taras and his code xD, secondly, it's his duty to make tests for that)

### DIFF
--- a/service/api_logic/auth_strategy.py
+++ b/service/api_logic/auth_strategy.py
@@ -114,12 +114,12 @@ class GoogleAuthHandler(AuthHandler[T]):
             output_login.username = new_user.username
             await self._user_service.create_tokens(user=output_login)
         else:
-            output_login.new_user = False   
+            output_login.new_user = False
+            output_login.user_id = user.user_id
+            output_login.username = user.username
         
         token = await self._user_service.get_generate_auth_token(user)
         
         output_login.token = token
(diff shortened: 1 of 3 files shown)


Avariq (approved):
Mostly alright

A couple suggestions for the future though:

image

1. We usually specify exactly what we test and what outcome is expected of that test. (If one test has multiple outcomes - that's different story and the outcome in name may be omitted)

2.  (thought it derives from Taras' changes, y'all made similar mistakes at some point):
In your test you struggle to mock JWT functionality, because it is strongly coupled with UserService. Basically you cannot properly test them separately. 
In order to prevent such things from happening we should follow DI principle as well as SOLID: we want every component of our program to be loosely coupled with others (as much as possible). In simple words: try to design your program's components like Lego pieces - we use many different pieces to build the final "idea". Each such piece is then easily tested/replaced etc etc.

So, in this case it would've been nice to have some JwtHelper (which I actually mentioned many times :) ) which would do all the token-related stuff. Then we simply use that helper wherever (inside user_service in this case). 

Make sure everybody else gets to read this, since it is quite an important app designing thing.

Approved nonetheless

PR #53, opened 2025-03-30
UJP-204 connect user favorite tems

Iarynovskyi:
IDK if i need to wait Vlads email sender but now it work and save correct users preference teams to db


Avariq (approved):
Seems alright (backendwise)

PR #57, opened 2025-04-04
Main page adaptation

Taras4568:
ujp-181 adjust main page


frontend/src/components/containers/gridBlock.jsx
```diff
@@ -18,6 +20,8 @@ const GridContainer = ({
     onPageChange,
     setSortValue
 }) => {
+    const noThirdButton = useBurgerMenu(1024);
+    const noSecondButton = useBurgerMenu(768);
```
AndrewKozovyi, line 24: convert these magic numbers to variables

frontend/src/pages/mainPage.js
```diff
@@ -23,14 +24,16 @@ function MainPage() {
     const { t } = useTranslations();
     const [news, setNews] = useState([])
     const [newsPaginated, setNewsPaginated] = useState([])
+    const noLatestNews = useBurgerMenu(`${globalVariables.windowSizeForBurger.latestNews}`);
```
Avariq, line 27: that's usually not a good idea to name your variables having them deny something (no... without...). It's better to go from positive, e.g. check if something EXISTS, and then, if not the '!<variableName>' makes much more sense and is easier to understand.

So in this case I'd go with "latestNews" or "latestNewsPresent"
something like that.

However, not that I look closer, this variable barely has something to do with news. It is focused on window size, is it not? 

Please rename this variable so that it reflects what it holds

frontend/src/pages/mainPage.js
```diff
@@ -23,14 +24,16 @@ function MainPage() {
...
+    const noLatestGames = useBurgerMenu(`${globalVariables.windowSizeForBurger.latestGames}`);
 
     const game_element_height = 85
     const game_element_width = 400
 
     const cardSizes = {
-        large: { rows: 1, columns: 4, cardSize: { width: 320, height: 490 }, postsPerPage: 4 },
-        medium: { rows: 3, columns: 5, cardSize: { width: 250, height: 300 }, postsPerPage: 10 },
-        small: { rows: 5, columns: 2, cardSize: { width: 650, height: 100 }, postsPerPage: 8 }
+        large: { rows: 1, columns: 4, cardSize: { width: 20, height: 30.625 }, postsPerPage: 4 },
```
Avariq, line 34: MINOR:
We usually do everything in % of the screen height/width. That way we have our UI automatically adjsut to most of the screens.

I didn't force y'all to do that in our project, because we had more than enough trouble with UI as is (and not too many volunteers as well :) ). So this one is just FYI for future

frontend/src/pages/mainPage.js
```diff
@@ -184,15 +187,15 @@ function MainPage() {
...
     };
 
     const [selectedSport, setSelectedSport] = useState("all");
-    const element_height = 100
-    const element_width = 350
     return (
         <>
             <div className="showcase">
-                <section>
+                {!noLatestNews ? (
```
Avariq, line 197: See? You basically get: "no no latest news" here

While not a mistake - not goodlooking either

Avariq (approved):
Feel free to merge after you rename that one variable

I haven't checked .cs files, but otherwise approved

frontend/src/pages/mainPage.js
Taras4568: Omg, you're checking js files? I can tell you a secret: I hardcoded this thing, 'cause the previous person wrote everything like this.

Taras4568: I only changed sizes from px to rem for better adaptation

PR #59, opened 2025-04-06
Deploy to Azure

Iarynovskyi:
We run docker-compose on Azure VM and our fiksik did not die
localhost front can work with deployed back, but problem with coolies setting and getting. We have different domens..
soon today "maybe..", we will try to fix this problem, but that will be difficult i think

Знімок екрана 2025-04-06 о 12 22 30


AndrewKozovyi (commented):
Срань

Iarynovskyi:
> Срань

Жопі слово не давали

AndrewKozovyi:
> > Срань
> 
> Жопі слово не давали

іди у дупу

